Issue #19, opened 2017-01-26
Add Sandbox Support


todofixthis:
Create a `SandboxAdapter` that enables developers to connect to remote nodes safely.

```
class SandboxAdapter(HttpAdapter):
  def configure(uri, token):
    # type: (Text, Text) -> SandboxAdapter
    ...

  def __init__(token, host, port=DEFAULT_PORT, path='/commands', poll_interval=15, sandbox_commands={'attachToTangle'}):
    # type: (Text, int, Text, int, Iterable[Text]) -> None
    ...
```

For non-sandbox commands:

1. Add header: `Authorization: token {self.token}`
2. Append the command name to the URI (e.g., `https://sandbox.iotatoken.com/api/v1/commands/attachToTangle`).
3. Return node response like normal.

For sandbox commands:

1. Add header: `Authorization: token {self.token}`
2. Append the command name to the URI (e.g., `https://sandbox.iotatoken.com/api/v1/commands/attachToTangle`).
3. Response will contain job info.  
  https://github.com/iotaledger/documentation/blob/sandbox/source/index.html.md#attachtotangle
4. Poll every `{self.interval}` to see if the job has completed.  
  https://github.com/iotaledger/documentation/blob/sandbox/source/index.html.md#jobs

todofixthis:
Scheduled for release: 1.1.0

Issue #13, opened 2017-01-14
udp is not a valid protocol for `HttpAdapter`


todofixthis:
tl;dr `HttpAdapter` does not actually send API requests via UDP, so it should not accept `udp://` URIs.

`udp://` URIs are only valid as neighbor URIs (i.e., as inputs for `addNeighbors` and `removeNeighbors`), because the node uses UDP for communication with other nodes.

The API, however, is an HTTP service; it doesn't use UDP, so `HttpAdapter` should not accept `udp://` URIs.

todofixthis:
Scheduled for release: 1.1.0

Issue #7, opened 2017-01-04
Support https node URIs 


todofixthis:
Although IRI only listens for API requests on the loopback interface, some developers opt to set up a reverse proxy.

Add support for connecting to a node via HTTPS.

todofixthis:
Scheduled for release: 1.1.0

Issue #6, opened 2017-01-04
Neighbor URIs must use udp protocol


todofixthis:
The IRI implementation only supports `udp://` URIs for neighbors.  If you send an `addNeighbors` API request containing an `http://` URI, that URI will be silently ignored.

When validating an `addNeighbors` or `removeNeighbors` request, treat any URIs that don't use `udp://` protocol as invalid.

todofixthis:
Will be delivered with #13

todofixthis:
Scheduled for release: 1.1.0

PR #21, opened 2017-01-28
1.1.0rc1

todofixthis:
**IMPORTANT:** This version includes several changes that are not backwards-compatible with v1.0.x!

# Backwards-Incompatible Changes
- Dropped support for IRI nodes below v1.1.2.4.
- [#13] Removed support for `udp://` API URLs.  Use `http://` instead.
- `getNewAddresses` now returns a dict. All API commands now have consistent return types.
- Restricted which commands each API class can create (`StrictIota` can only create core commands; `Iota` can create core+extended).
   - `__getattr__` now raises an `InvalidCommand` error if an unsupported command is requested (previously it would return a `CustomCommand` instance).
   - Implemented `create_command` method to create `CustomCommand` objects.

# New Features
- [#19] Implemented support for sandbox nodes.

# Enhancements/Fixes
- [#6] `addNeighbors` and `removeNeighbors` now only accept `udp://` URIs.
- [#7] Added support for `https://` API URLs.
- Improved handling of non-200 responses from nodes.
- Adapters now use `urlsplit` to parse URIs.
- Adapters now provide more context vars when raising `BadApiResponse`.
- `discover_commands` now also returns all command objects that it finds.
- Changed `requests` dependency to `requests[security]`.
  - See https://github.com/kennethreitz/requests/issues/1995 for more info.
- Cleaned up code in `examples/shell.py`.
- General documentation cleanup and fixes.

### DIFF
--- a/examples/sandbox.py
+++ b/examples/sandbox.py
@@ -1,0 +1,55 @@
+# coding=utf-8
+from __future__ import absolute_import, division, print_function, \
+  unicode_literals
+
+from iota import *
+from iota.adapter.sandbox import SandboxAdapter
+
+
+# Create the API object.
+iota =\
+  Iota(
+    # To use sandbox mode, inject a ``SandboxAdapter``.
+    adapter = SandboxAdapter(
+      # URI of the sandbox node.
+      uri = 'https://sandbox.iotatoken.com/api/v1/',
+
+      # Access token used to authenticate requests.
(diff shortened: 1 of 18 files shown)


todofixthis:
Now that 1.0 is released, we should avoid merging pre-release versions into `master`.  Create a release branch instead.

Issue #18, opened 2017-01-24
Add unit converter for transaction values


todofixthis:
IOTA utilizes the Standard system of Units. See below for all available units:

```
'i'   :   1,
'Ki'  :   1000,
'Mi'  :   1000000,
'Gi'  :   1000000000,
'Ti'  :   1000000000000,
'Pi'  :   1000000000000000
```

Add support for specifying units in transactions.  E.g.:

```python
iota.send_transfer(
  transactions = [
    ProposedTransaction(
      value = '86.2 Mi',
      ...
    ),
    ...
  ],
  ...
)
```

jinankjain:
@todofixthis I am looking forward to work on this bug.
I am thinking of making a separate function in transactions.py which will convert the value from string to int. Something like this:

```python
def convert_value_to_standard_system_units(value):
  # type: string -> float
  """
  Return a floating point value in Standard System of Units from string
  'i'   :   1,
  'Ki'  :   1000,
  'Mi'  :   1000000,
  'Gi'  :   1000000000,
  'Ti'  :   1000000000000,
  'Pi'  :   1000000000000000
  """
  value_tuple = value.split()
  unit = 0
  if(value_tuple[1] == "i"):
    unit = 1
  elif(value_tuple[1] == "Ki"):
    unit = 1000
  elif(value_tuple[1] == "Mi"):
    unit = 1000000
  elif(value_tuple[1] == "Gi"):
    unit = 1000000000
  elif(value_tuple[1] == "Ti"):
    unit = 1000000000000
  elif(value_tuple[1] == "Pi"):
    unit = 1000000000000000
  
  return float(value_tuple[0])*unit
```

Is this the correct approach?

todofixthis:
Hey @jinankjain.  The approach you've laid out looks good.  I think it could be improved with some validation (`value` must have correct type and format, must match a recognized value).

Looking forward to your pull request (:

vynes:
Hi!

I've had a crack at it.
What return value is expected if validation fails? I've returned _False_.
I've also set the default unit prefix ('i') if none given. Not 100% sure if such implicit setting could cause problems?
Those three WARNING/ERROR print() statements are for testing purposes and should probably be replaced by logger methods, if such an approach will be used? (else just removed.)

```Python
def convert_value_to_standard_system_units(value):
    # type: string -> float
    """
    Return a floating point value in Standard System of Units from string
    """
    standard_units = {
        'i': 1,
        'Ki': 1000,
[...]

PR #48, opened 2017-05-10
[#18] Add unit converter

todofixthis:


test/api_test.py
```diff
@@ -135,3 +135,14 @@ def test_create_command(self):
...
+  def test_convert_units(self):
+    """
+    Converting between iota standard units.
+    """
+    api = Iota(MockAdapter())
+
+    self.assertEqual(api.convert_units('1.618 Mi', 'i'), 1618000)
+    self.assertEqual(api.convert_units('42 i', 'Ki'), 0.042)
+    with self.assertRaises(ValueError):
+      api.convert_units('3.141596 Xi', 'Pi')
```
todofixthis: `convert_value_to_standard_unit` deserves its own test case.  This makes it easier to test just that function in isolation from the rest of the library.

Also, I would like each scenario split into its own test (e.g., create `test_convert_to_smaller_unit`, `test_convert_to_bigger_unit`, `test_error_invalid_unit`, and so on).  This makes it easier to understand what is being tested, and if any assertion fails, it doesn't prevent the other assertions from being checked.

Lastly, there are a few more tests that need to be added for full coverage:

- Input is the wrong type (e.g., `[1.618, 'Mi']`).
- Input is already a float value (e.g., `1.618`).  Probably same result as the previous one.
- Input is empty (i.e., `''`).
- Input has too few parts (e.g., `'1.618`).
- Input is missing space (e.g., `'1.618Mi'`).  We can treat this as an error.
- Input has too many parts (e.g., `'1 618 Mi'`).
- Result includes fractional IOTAS (e.g., `'1.6189 Ki'`).  This is not an error, just using unit tests to document that this behavior is allowed.
- Input is negative (e.g., `'-1.618 Mi'`).
- Input uses different case (e.g., `'1.618 mI'`).  This should probably be an error... probably.

src/iota/transaction.py
```diff
@@ -45,6 +46,49 @@ def get_current_timestamp():
...
+
+    :return:
+      float as units of symbol to convert to
+  """
+  value_tuple = value.split()
+
+  try:
+    # Get input value
+    amount = float(value_tuple[0])
+  except (ValueError, IndexError) as e:
```
todofixthis: Is `e` being used here?  Double-check your IDE's configuration; it should highlight this as an issue.

src/iota/__init__.py
```diff
@@ -17,6 +17,17 @@
 In that way, it's kind of like toxic waste in a superhero story.
 """
 
+STANDARD_UNITS = {
+    # Valid IOTA unit prefixes. Example value '25.2 Mi'
```
todofixthis: ... and *that's* the first thing that appears in this PR.  Great; put my pedantry on display front and center :P

test/api_test.py
curVV: I've created a test case `class UnitConverterTestCase(TestCase)` in `test/transaction_test.py`
How do run only this test case?
Can it be done using setup.py?

curVV: OK, I've managed to at least run all tests in `transaction.py` with `python setup.py test -s test.transaction_test`
Not getting it to run `test.transaction_test.UnitConverterTestCase` only, though. Unit test runs but reports: `Ran 0 tests in 0.001s`

todofixthis:
Changes look good!  Thanks!

PR #50, opened 2017-05-20
Fix issue in address_generator example.

ivoscc:
The `get_new_addresses` method returns a dict and we were iterating over it's
keys.


todofixthis:
jej oops (:  Thanks Ivo!

Issue #55, opened 2017-07-21
Incorrect attribute names in GetBalancesCommand response (using sandbox example script)


johngrantuk:
PyOTA 1.1.3

Python 2.7.10

Mac OS El Capitan 10.11.6

**Description:**
Running sandbox [example](https://github.com/iotaledger/iota.lib.py/blob/master/examples/sandbox.py). iota.send_transfer gives following error: TypeError: 'NoneType' object is not iterable

**Traceback:**
Traceback (most recent call last):
  File "sandbox.py", line 51, in <module>
    message = TryteString.from_string('Hello, Tangle!'),
  File "/Users/johngrant/Downloads/iota/env/lib/python2.7/site-packages/iota/api.py", line 822, in send_transfer
    minWeightMagnitude  = min_weight_magnitude,
  File "/Users/johngrant/Downloads/iota/env/lib/python2.7/site-packages/iota/commands/__init__.py", line 122, in __call__
    self.response = self._execute(self.request)
  File "/Users/johngrant/Downloads/iota/env/lib/python2.7/site-packages/iota/commands/extended/send_transfer.py", line 46, in _execute
    transfers       = transfers,
  File "/Users/johngrant/Downloads/iota/env/lib/python2.7/site-packages/iota/commands/__init__.py", line 122, in __call__
    self.response = self._execute(self.request)
  File "/Users/johngrant/Downloads/iota/env/lib/python2.7/site-packages/iota/commands/extended/prepare_transfer.py", line 54, in _execute
    threshold = want_to_spend,
  File "/Users/johngrant/Downloads/iota/env/lib/python2.7/site-packages/iota/commands/__init__.py", line 122, in __call__
    self.response = self._execute(self.request)
  File "/Users/johngrant/Downloads/iota/env/lib/python2.7/site-packages/iota/commands/extended/get_inputs.py", line 62, in _execute
    for i, balance in enumerate(gb_response['balances']):
TypeError: 'NoneType' object is not iterable

**Possible fix? (Creating a Pull request)**
In get_inputs.py, Line 51, gb_response = GetBalancesCommand( ) is returning: {u'MilestoneIndex': 123423, …. u'Duration': 0, u'Balances': [0, 0, 41, 0, 0, 0]}

Line 62 uses: enumerate(gb_response['balances'])

Changing 'balances' to 'Balances' seems to fix the problem.

todofixthis:
Hi @johngrantuk!  Thanks for reporting this issue!

[...]

PR #56, opened 2017-07-21
Changed 'balances' to fix Issue #55 get_inputs.py Issue (Using Sandbox)

johngrantuk:
A suggested change that fixes the issue on my machine. 
Hopefully I've done the process the correct way, first time trying. Thanks.

### DIFF
--- a/iota/commands/extended/get_inputs.py
+++ b/iota/commands/extended/get_inputs.py
@@ -50,7 +50,7 @@ class GetInputsCommand(FilterCommand):
       # Load balances for the addresses that we generated.
       gb_response = GetBalancesCommand(self.adapter)(addresses=addresses)
     else:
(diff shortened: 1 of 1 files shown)


todofixthis:
Hey @johngrantuk!  Thank you for your pull request!

This issue is most likely caused by a problem with the node itself — the attribute names actually _should_ be lowercase.

If necessary, we might have to add some compatibility middleware that fixes incorrect attribute names (but we would need to take a more holistic approach to ensure that _every_ attribute name is fixed, not just `balance`).  I'm hoping we don't have to do that, though; better if the node maintainer fixes their code (:

When you have a chance, please reply to my comment in #55 with the raw response payload from the `getBalances` command, so that I can follow up with the sandbox node maintainer to see what's going on.

Thanks!!

Issue #31, opened 2017-03-05
Example script for creating transactions


todofixthis:
Write an example script that shows how to use PyOTA to send a transfer to another address.

todofixthis:
Scheduled for release: 1.2.0
Thanks @johngrantuk!

PR #58, opened 2017-07-30
Attempt at Issue #31.

johngrantuk:
Attempt at an example script that shows how to use PyOTA to send a transfer to another address.


todofixthis:
Thanks!  Merging..

PR #71, opened 2017-09-24
[#67] Attachment Timestamps

### LINKED ISSUE #67: Attachment Timestamps
In a recent snapshot, the structure of IOTA transactions has changed a bit:

- Rename `tag` to `legacy_tag`.
- Add `tag`, `attachment_timestamp` , `attachment_timestamp_lower_bound` and `attachment_timestamp_upper_bound` after `branch_transaction`.
- Resize `nonce` from 81 to 27 trytes.

More info:
- https://github.com/iotaledger/iri/pull/303
- https://github.com/iotaledger/iota.lib.js/pull/79/files

normpad:
Fixes #67 

* Updated transactions base to new structure
* Updated transaction creation to new structure 
* Updated affected test cases

Let me know if I missed anything.


iota/transaction/base.py
```diff
@@ -69,16 +73,20 @@ def __init__(
       signature_message_fragment,
       address,
       value,
-      tag,
+      legacy_tag,
```
todofixthis: In a future snapshot (probably the next one), `legacy_tag` will be removed entirely and replaced with the new `tag`.

For now, let's just treat `legacy_tag` as an alias to new `tag`:

- Initialiser accepts `tag` only (no `legacy_tag`).
- Set `self.tag` as normal.
- Make `self.legacy_tag` an alias for `self.tag` (recommend using a `@property` for that, just in case).

iota/transaction/base.py
```diff
@@ -104,9 +112,9 @@ def __init__(
...
     """
 
-    self.tag = tag # type: Optional[Tag]
+    self.legacy_tag = legacy_tag # type: Optional[Tag]
     """
-    Optional classification tag applied to this transaction.
+    Optional classification legacy_tag applied to this transaction.
     """
 
     self.nonce = nonce # type: Optional[Hash]
```
todofixthis: This one may need some love as well — `Hash` automatically pads itself to 81 trytes.

iota/transaction/base.py
normpad: I left legacy_tag in the initializer defaulted to None. The reason being when creating the transaction from the tryte string legacy_tag is extracted and needs a way to be set. It gets set that way in the JS library so I figured it would be best to keep it consistent. If legacy_tag is not set it will just point to the tag.

iota/transaction/types.py
```diff
@@ -72,3 +73,25 @@ def __init__(self, trytes):
...
+      raise with_context(
+        exc = ValueError('{cls} values must be {len} trytes long.'.format(
+          cls = type(self).__name__,
+          len = self.LEN
+        )),
+
+        context = {
+          'trytes': trytes,
+        },
+      )
```
todofixthis: Ooh, good idea!

todofixthis:
Thanks for your updates!  One last thing; could you take a look at the unit test failures?
https://travis-ci.org/iotaledger/iota.lib.py/jobs/279143972

normpad:
Yea I overlooked some of them. Sorry! Fixed now.

todofixthis:
Gracias amigo!  Taking a look now.

test/commands/extended/get_bundles_test.py
```diff
@@ -353,7 +356,7 @@ def test_multiple_transactions(self):
           b'DRXICGYDGSVPXFTILFFGAPICYHGGJ9OHXINFX9999'
         ),
     )
-
+    print(response['bundles'][0].as_tryte_strings())
```
todofixthis: Oops (:

normpad: Yea I had to do some investigating on that one...

todofixthis:
I'm going to give this a final once-over tonight; need to validate it against the JS library.

todofixthis:
Looks good!  Merging....

Issue #43, opened 2017-04-25
Set up private Tangle for functional testing


todofixthis:
Many of PyOTA's unit test rely on heavy use of mocks, which reduces the value of the test coverage significantly.

A more robust approach would be to set up automated functional tests, on a private Tangle.  Per conversation with Paul H on Slack, we should be able to set this up without too much difficulty:

1. Clone the repo.
2. Check out the `testnet` branch.
3. Modify the `Snapshot.java` file to configure a custom genesis block.
4. `mvn clean compile package`
5. Run the node with `--testnet` flag.
6. ???
7. Profit!!!

Set up a proof of concept with a private Tangle and a script that transfers IOTAs between two addresses successfully.  We will iterate on it from there and create a comprehensive functional test suite 😺

todofixthis:
The min weight magnitude is set here:
https://github.com/iotaledger/iri/blob/testnet/src/main/java/com/iota/iri/TransactionValidator.java#L19

It would be nice if we could ratchet this down for functional tests, but it may be tricky to modify this cleanly.

jinnerbichler:
This can be done via Docker containers (e.g. https://docs.travis-ci.com/user/docker/). 
Would that be a feasible approach?

todofixthis:
Aye, I think that would be a good way to go about it.  I was talking with the folks at Travis CI; here's the outcome of the conversation:
 
1. (optimisation) Use build stages so that the unit tests run first, and then the functional tests run only if the unit tests passed.
2. Add `services: docker` to `.travis.yml` (https://docs.travis-ci.com/user/docker/).
3. Add a `before_install` step that downloads+modifies the IRI source and builds the Docker image.
4. Add to `script`:
    - Use `docker` command to launch 3 Docker containers using the image built in the previous  step.
    - Run functional tests.

huntal:
So does the research advanced to some point ?

todofixthis:
Hey Huntal.  Yeah, this guy's been pending for a really long time ):

I'm actually working on something similar at my job right now ("Instant Testnet:  Just add Docker!"), so I might be able to make some real progress on this once I hammer out all of the kinks.

That said, if anyone wants to take this on, just let me know; I'd be happy to collaborate!

huntal:
I'll be happy to work with you on this project but i may think i'm less strong than you in coding master PyIOTA ;)

todofixthis:
Thanks Hystenal!

[...]

PR #73, opened 2017-09-29
Fixes replayed bundle being reversed

### LINKED ISSUE #74: replayBundle sends trytes in reverse order
When invoking the `replayBundle` command, the resulting trytes are broadcast in reverse order, preventing `getBundles` from working correctly.

Example:  
https://thetangle.org/bundle/GMMGJTEWAINANUTNRNQLUEZFTTTSEPODSODQFJWCPGBMFZHXL9ZLJXBDMDZKCKERCOZRRCXCP9AQLIG9X

Bundle should look like this (note reverse order):  
https://thetangle.org/bundle/EYTSBAEABMXUNDVOMXBGPZWVGXTXGYHENBAMEETOSHXMNXUBSTUOUXOLVOFRCBXBZXIOMVWRPSYYEFEVX

normpad:
Fixes #74


todofixthis:
Hey @normpad thanks for submitting this pull request!

Could you include a unit test for regressions (i.e., if somebody later goes back and reverses the order of transactions again, the unit test should fail)?

The fact that it took us so long to catch this bug manually makes me concerned; I would sleep a lot better at night if we had an automated test for it (:

normpad:
Absolutely, I'm just not entirely sure how to go about that. But assuming trytes should always be passed to attachToTangle tail to head, wouldn't it be fitting to have a check before passing the trytes to the node and throw an error if they are out of order?

todofixthis:
Hm.  Interesting.

Here's what I think you can do:

Modify this line:  `test/commands/extended/replay_bundle_test.py:599`
```python
mock_send_trytes = mock.Mock(return_value=send_trytes_response)
```

Change it to something like this:
```python
def mock_send_trytes(request):
  """
  Ensures that the correct trytes are sent to the ``sendTrytes`` command
  to replay the bundle.

  References:
    - https://github.com/iotaledger/iota.lib.py/issues/74
  """
  self.assertEqual(
    request['trytes'],

    [
      ... transaction trytes go here ...
    ]
  )

  return send_trytes_response

...

  with mock.patch(
      'iota.commands.extended.send_trytes.SendTrytesCommand._execute',
      mock_send_trytes,
  ):
    ...
```

todofixthis:
Incidentally, once https://github.com/iotaledger/iota.lib.py/issues/43 is implemented, we can set up a real private subtangle when running unit tests, and then we can get rid of all of this mocking nonsense 😸

test/commands/extended/replay_bundle_test.py
```diff
@@ -593,10 +593,20 @@ def test_happy_path(self):
...
+    def mock_send_trytes(self,request):
+      """
+      Ensures that the correct trytes are sent to the ``sendTrytes`` command
+      to replay the bundle.
+
+      References:
+        - https://github.com/iotaledger/iota.lib.py/issues/74
+      """
+      from unittest import TestCase
+      TestCase().assertEqual(request['trytes'], send_trytes_response['trytes'])
```
todofixthis: Python 2.7 doesn't like that so much.  Maybe something like this:

```python
def mock_send_trytes(_, request):
  ...
  self.assertEqual(...)
```

todofixthis: Aye.  I admit; it feels like a hack, either way, but that's why I'm so keen on getting #43 set up (:

todofixthis:
Looks great!  Merging now.

As far as I'm aware, this was the last obstacle to releasing v2.0.0 stable (:  Thanks!!!

PR #75, opened 2017-09-29
fix transfer = ProposedTransaction(...) naming convention doesn't follow other libs #72

### LINKED ISSUE #72: transfer = ProposedTransaction(...) naming convention doesn't follow other libs
both in `iota.lib.js` and `iota.lib.java` there are explicit `transfer` objects.
[iota.lib.js](https://github.com/iotaledger/iota.lib.js/blob/master/lib/api/api.js#L1024)
[iota.lib.java](https://github.com/iotaledger/iota.lib.java/blob/master/src/main/java/jota/IotaAPI.java#L366)

however, in `iota.lib.py` these objects are constructed with `iota.ProposedTransaction(...)`, which is also used as an intermediary representation of [transactions](https://github.com/iotaledger/iota.lib.py/blob/master/iota/transaction/creation.py#L28) & bundles.
from examples:
https://github.com/iotaledger/iota.lib.py/blob/master/examples/send_transfer.py#L24

this is confusing for users moving from other library implementations. like myself. :)

---
I propose a new class named `Transfer`, that inherits from `ProposedTransaction` and is functionally the same, but adheres to the naming conventions of the other libs. (without breaking code that already uses `ProposedTransaction` of course).

szepnapot:
Fixes #72

### DIFF
--- a/iota/transaction/creation.py
+++ b/iota/transaction/creation.py
@@ -80,6 +80,42 @@ class ProposedTransaction(Transaction):
     return super(ProposedTransaction, self).as_tryte_string()
 
 
+class Transfer(ProposedTransaction):
+  """
+  Follow naming convention of other libs.
+  A transaction that has not yet been attached to the Tangle.
+
+  Provide to :py:meth:`iota.api.Iota.send_transfer` to attach to
+  tangle and publish/store.
+  """
+  def __init__(self, address, value, tag=None, message=None, timestamp=None):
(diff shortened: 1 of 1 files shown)


todofixthis:
Hey @szepnapot!  Thanks for your pull request!

I will have a look at this later today.

szepnapot:
True! I'll modify according to it. Thanks the review :)

szepnapot:
Changed according the suggestion :)

todofixthis:
Thanks!  I'll have a look this evening.

Issue #45, opened 2017-04-29
Seed.__init__ should raise a warning when the seed is longer than 81 trytes


todofixthis:
Beyond 81 trytes, you can't increase the security of your seed — not even against brute force attacks (see [Why aren't seeds longer than 81 trytes more secure?](https://forum.iota.org/t/why-arent-seeds-longer-than-81-trytes-more-secure/1278/) for more info).

We should still allow users to create seeds longer than 81 trytes, just in case, but this should cause a warning to be raised.  The error message should include a link to the above forum post.

Example:
```python
# 81 trytes; this is OK.
seed = Seed(b'9' * Hash.LEN)

# 82 trytes; this should raise a warning (note: not an exception)
seed = Seed(b'9' * (Hash.LEN + 1))

# If the user is sure they know what they are doing, they can suppress the warning.
with warnings.catch_warnings():
  warnings.simplefilter('ignore')
  seed = Seed(b'9' * (Hash.LEN + 1))
```

Note:  Do not strip trailing `9`s when checking the length.

todofixthis:
Note: coordinate this with #44

todofixthis:
Scheduled for release: v2.0.1

PR #77, opened 2017-10-06
Issue #45: Seed check

jinnerbichler:
fixing issue #45


jinnerbichler:
There might be something wrong with the CI.
https://travis-ci.org/iotaledger/iota.lib.py/jobs/284322913

todofixthis:
Hi Johannes.  Thanks for your PR!

Aye, looks like there was a temporary hiccup with Travis CI; I restarted the build and it succeeded.

I'll have a look at this in a little bit.

iota/crypto/types.py
```diff
@@ -76,6 +77,15 @@ class Seed(TryteString):
   References:
     - https://forum.iota.org/t/why-arent-seeds-longer-than-81-trytes-more-secure/1278
   """
+
+  def __init__(self, trytes=None):
+    # type: (Optional[TrytesCompatible]) -> None
+    if trytes and len(trytes) > 81:
```
todofixthis: This magic number can be replaced with `Hash.LEN`.

.gitignore
```diff
@@ -19,3 +19,5 @@ dist/*
 # you can add these to your own "global" .gitignore file.
 # https://help.github.com/articles/ignoring-files/#create-a-global-gitignore
 #
+.idea/
+.python-version
```
todofixthis: These can go in your "global" `.gitignore` file, as they're not specific to PyOTA (see https://help.github.com/articles/ignoring-files/#create-a-global-gitignore).

BTW, nice to see another PyCharm user!

test/crypto/signing_test.py
```diff
@@ -34,144 +35,151 @@ def test_get_keys_single(self):
...
     Generating a single key.
     """
-    kg = KeyGenerator(
-      seed = b'TESTSEED9DONTUSEINPRODUCTION99999ZTRFNBTRBSDIHWKOWCFBOQYQTENWL',
-    )
+    with warnings.catch_warnings(record=True) as catched_warnings:
 
-    self.assertListEqual(
-      kg.get_keys(start=0),
+      warnings.simplefilter("always")
```
todofixthis: Can this be made more specific, so that it only catches the "Seed has invalid length" warning?

Just in case a regression is introduced in the future that breaks the seed length warning and coincidentally causes an unrelated warning to be emitted (e.g., we add a deprecation warning somewhere, and we forget to update the `Seed` class to account for whatever got changed).

jinnerbichler:
Hi @todofixthis 
totally makes sense.
I have introduced a seed specific warning and updated the PR.

todofixthis:
Thanks!  I really appreciate it!

Could you change the `assert` statements to TestCase assert methods (e.g., `self.assertEqual()`)?

Then I think this should be good to go!

todofixthis:
Thanks!  Doing a few final checks before merging.

todofixthis:
Looks good!  Merged.  Thanks!!

Issue #62, opened 2017-08-25
TryteString.as_bytes() does not work as intended


th0br0:
Expected result (or similar):
```
>>> iota.TryteString(b'ABABZZDD99').as_bytes()
(30, [55,178,248,107,12,0])
```

Actual result:
```
>>> iota.TryteString(b'ABABZZDD99').as_bytes()
Traceback (most recent call last):
  File "iota.lib.py/iota/codecs.py", line 147, in decode
    + (self.index[second] * len(self.index))
ValueError: byte must be in range(0, 256)

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "iota.lib.py/iota/types.py", line 421, in as_bytes
    return decode(self._trytes, 'trytes', errors)
  File "iota.lib.py/iota/codecs.py", line 165, in decode
    'input': input,
iota.codecs.TrytesDecodeError: 'trytes' codec can't decode trytes ZZ at position 4-5: ordinal not in range(255)
```

todofixthis:
Hey @th0br0.  This is actually expected behaviour — the tryte sequence `ZZ` represents value -28, which cannot be converted into a single byte.

In order to convert this sequence into a byte string, you must instruct PyOTA to handle invalid sequences differently (this is very similar to how `string.decode('utf-8')` handles invalid bytes):

```python
>>> from iota import TryteString

>>> TryteString(b'ABABZZDD99').as_bytes('strict')
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/phx/Documents/pyota/iota/types.py", line 498, in as_bytes
    return decode(self._trytes, 'trytes', errors)
  File "/Users/phx/Documents/pyota/iota/codecs.py", line 165, in decode
    'input': input,
iota.codecs.TrytesDecodeError: 'trytes' codec can't decode trytes ZZ at position 4-5: ordinal not in range(255)
'trytes' codec can't decode trytes ZZ at position 4-5: ordinal not in range(255)

>>> TryteString(b'ABABZZDD99').as_bytes('replace')
b'77?p\x00'

>>> TryteString(b'ABABZZDD99').as_bytes('ignore')
b'77p\x00'
```

th0br0:
I beg to differ. I'm not well-versed enough in Python3 codecs to pinpoint exactly where things need to change, but:
 `as_bytes()` should split the trits into chunks of up to 5 trits each and then essentially do:
```python
[...]

PR #88, opened 2017-10-31
Add TryteString to bytes test cases

mlouielu:
As for issue #62, this is the test cases for converting
TryteString to bytes, and bytes to TryteString.

(CI dead is correct)

### DIFF
--- a/test/types_test.py
+++ b/test/types_test.py
@@ -7,7 +7,194 @@ from unittest import TestCase
 from six import binary_type, text_type
 
 from iota import Address, AddressChecksum, AsciiTrytesCodec, Hash, Tag, \
-  TryteString, TrytesDecodeError
+  TryteString, TrytesDecodeError, TransactionTrytes, TransactionHash
+
+
+# noinspection SpellCheckingInspection
+class TryteStringConvertBytesTestCase(TestCase):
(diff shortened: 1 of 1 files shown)


todofixthis:
Hey @mlouielu thank you for working on this!  I've asked some of the core devs (who are better at maths than I am 😸) to have a look at this.

I had a few questions about these tests:
- What should `TryteString('NOP').as_bytes()` return?  (negative value)
- I see that `TryteString.from_bytes('', length=1)` returns `TryteString('9')`.  But, `TryteString('').as_bytes(length=1)` raises an `IndexError`; wouldn't it be more consistent if it returned `b'\x00'`?
- I think the tests cover this, but help me to understand because I still don't get this 😇 :
    - Converting 27 trits to bytes:  `27 % 5 != 0`, so how is this conversion handled?

todofixthis:
Hey @mlouielu thanks again for your help with this feature!  I'm working with a few individuals to try to nail down requirements for this codec — I feel like there are a number of edge cases that the existing implementations are not able to cover (https://github.com/vbakke/trytes/ has an excellent write-up of many of these issues).

In the meantime, I've put #62 on hold until we can work out all of the details.

Issue #92, opened 2017-11-06
Add offset and length argument to pycurl absorb and squeeze


mlouielu:
Currently, pycurl absorb and squeeze only have `trits` argument, but `pykerl` have offset and length.

Although pycurl is a deprecated hash function, but in IRI snapshot, it used `CURLP81` to sign the snapshot, so if we want to verify snapshot signature, we will need offset and length argument to used `absorb` and `squeeze`.

todofixthis:
Hey @mlouielu thanks for reporting this!

Looks fairly straightforward to implement this; I'm going to tag this issue with `help wanted`.

todofixthis:
Scheduled for release: 2.0.4

PR #95, opened 2017-11-09
Add offset and length to pycurl

mlouielu:
Issue: #92 

pycurl.Curl.absorb:
    Add offset and length

pycurl.Curl.squeeze:
    Add offset

    sequeeze not added with length, is because it will only
    copy one hash to trits, therefore no need to add length


todofixthis:
Thanks @mlouielu !  I will take a look at this over the weekend.

iota/crypto/pycurl.py
```diff
@@ -59,16 +61,35 @@ def reset(self):
...
+
+    if length < 1:
+      raise with_context(
+        exc = ValueError('Invalid length passed to ``absorb``.'),
+        context = {
+          'trits': trits,
+          'offset': offset,
+          'length': length,
+        },
+      )
```
todofixthis: :+1:

iota/crypto/pycurl.py
```diff
@@ -92,14 +113,17 @@ def absorb(self, trits):
       # Move on to the next hash.
       offset += HASH_LENGTH
 
-  def squeeze(self, trits):
+  def squeeze(self, trits, offset=0):
```
todofixthis: Should `squeeze` also have a `length` argument?

In practice, it seems that this argument is always `HASH_LENGTH`, but the other libraries do support it.  Examples:

- https://github.com/iotaledger/iota.lib.java/blob/v0.9.6/src/main/java/jota/pow/JCurl.java#L156
- https://github.com/iotaledger/iota.lib.js/blob/v0.4.4/lib/crypto/curl/curl.js#L78

test/crypto/pycurl_test.py
```diff
@@ -0,0 +1,107 @@
+# coding=utf-8
+from __future__ import absolute_import, division, print_function, \
+  unicode_literals
+
+from unittest import TestCase
+
+from iota import TryteString
+from iota.crypto.pycurl import Curl
```
todofixthis: Could you change this to `from iota.crypto import Curl`?  That way, I can repurpose these tests for the C extension.

More info:
- https://github.com/iotaledger/iota.lib.py/blob/2.0.2/iota/crypto/__init__.py#L6-L13
- https://github.com/todofixthis/pyota-ccurl/blob/1.0.9/.travis.yml#L12-L17

test/crypto/pycurl_test.py
```diff
@@ -0,0 +1,107 @@
...
+
+
+class TestCurl(TestCase):
+    def test_correct_first(self):
```
todofixthis: Could you add docstrings to each unit test, describing the situation it is testing?  The test names are a good start, but a little extra documentation describing the context for each test will go a long way.

test/crypto/pycurl_test.py
```diff
@@ -0,0 +1,107 @@
...
+
+
+        d = [0] * 243
+        trits = TryteString(inp).as_trits()
+        curl = Curl()
+
+        for i in range(6):
+            curl.reset()
+            curl.absorb(trits, i * 243, (i + 1) * 243)
+            curl.squeeze(trits, i * 243)
```
todofixthis: This test looks very specific.  Is it mimicking a real-world use case?

Unit tests also serve as documentation; if you add some comments explaining what real-world situation this test is simulating, it will help users to better understand what they can accomplish with the library.

todofixthis:
Hey @mlouielu I feel like this PR is almost ready to merge.  I've requested a few changes, but they're mostly minor stuff, documentation and such.  Let me know if you have any questions, or if I can provide additional info.

mlouielu:
I'll find some time these days to make sure this PR fits your guide.

todofixthis:
Thanks Louie!

iota/crypto/pycurl.py
mlouielu: Added, but in real case, this doesn't use without 243 (HASH_LENGTH)

test/crypto/pycurl_test.py
mlouielu: This only use in ISS and ISSInplace, add the source code link in docstring.

todofixthis:
Hey Louie!  Thanks for working on this!  I will take a look this weekend and follow up.

Issue #78, opened 2017-10-14
Port documentation to Sphinx


todofixthis:
We need your help to port the existing documentation from Markdown to reStructuredText so that we can publish it on [ReadTheDocs](http://pyota.readthedocs.io/en/develop/)!

The Markdown files are located here:
https://github.com/iotaledger/documentation/tree/iota.lib.py/1.2.x/source/includes

These need to be copied into the PyOTA repository and converted to reStructuredText here:
https://github.com/iotaledger/iota.lib.py/tree/develop/docs

Note:  This is a big project, and multiple people might end up working on this.  Please be sure to post here so that you can coordinate with others, to avoid duplicating work.

If you are not familiar with Sphinx and/or reStructuredText, here's a quickstart:

1. Install PyOTA locally.
2. `pip install sphinx`
3. `cd docs`
4. `make html`
    * This will generate the static HTML documentation.  You can then open `docs/_build/html/index.html` in your browser to see what it will (sort of) look like when it is published to RTD.
5. To make changes to the documentation, edit the corresponding `.rst` file, and then do another `make html`.

**Important:** If you add a new `.rst` file, be sure to update the `toctree` directive in `index.rst`!

Here are some references for reStructuredText directives and markup syntax (once you get used to it, you will never want to go back to Markdown 😸):
- http://docutils.sourceforge.net/docs/user/rst/quickref.html
- http://docutils.sourceforge.net/docs/ref/rst/directives.html
- http://docutils.sourceforge.net/docs/ref/rst/restructuredtext.html

You can also use this online rst editor to try stuff out:
http://rst.ninjs.org/

QuantumDamage:
I'm starting to work on this issue.

QuantumDamage:
@todofixthis I see that there is already **types.rst** doc file. Should I overwrite it with content from **_types.md** file?

todofixthis:
@QuantumDamage Affirmative.  It might be partially-complete using `_types.md` as the source, so you might be able to leave some/all of what's already there.

QuantumDamage:
@todofixthis OK.

QuantumDamage:
#102

todofixthis:
Scheduled for release: 2.0.3

Thank you!!!

PR #102, opened 2017-12-01
Ported docs from md to rst #78

QuantumDamage:
I have ported **adapters**, **addresses** and **api** using pandoc and fixed by hand some errors. I also replaced old **types.rst** with freshly generated from **_types.md**. It builds locally.


todofixthis:
This is awesome @QuantumDamage thank you so much!!

Issue #89, opened 2017-10-31
Iota.get_new_addresses() function doesn't have a checksum option


danboard-7050:
JS API's get_new_addresses() function has a checksum option, but Python APIs' doesn't have it.

mlouielu:
Why will you need to use checksum option? After get_new_address(), it will return a `iota.Address`, if you want to get the address with checksum, use `iota.Address.with_valid_checksum()`

danboard-7050:
I see. Thanks.

todofixthis:
I'm going to go ahead and re-open this.  A lot of users have been asking about it, and we should try to make the Python lib resemble the JS lib, at least a little bit (:

todofixthis:
Scheduled for release: 2.0.3

PR #106, opened 2017-12-08
Issue #89 - Initial work on adding checksum option to get_new_addresses

plenarius:
Okay first shot at contributing. Still getting my feet wet with the API but this looked like something to get started. My attempt at Issue #89 

Should I have added the --security flag in the example address_generator.py in another commit?

The FindTransactions part with me having to chop the checksum back off, feels like that's pretty hackish.  And I wasn't entirely sure where the best place to put the with_valid_checksum() so I settled on in the _generate_address().

Anyway it's a start! #

### DIFF
--- a/examples/address_generator.py
+++ b/examples/address_generator.py
@@ -16,8 +16,8 @@ from iota.crypto.types import Seed
 from six import binary_type, moves as compat, text_type
 
 
-def main(uri, index, count):
-  # type: (Text, int, Optional[int], bool) -> None
+def main(uri, index, count, security, checksum):
+  # type: (Text, int, Optional[int], Optional[int], Optional[bool]) -> None
   seed = get_seed()
 
   # Create the API instance.
@@ -34,7 +34,7 @@ def main(uri, index, count):
   print('')
 
   # Here's where all the magic happens!
-  api_response = api.get_new_addresses(index, count)
+  api_response = api.get_new_addresses(index, count, security, checksum)
   for addy in api_response['addresses']:
     print(binary_type(addy).decode('ascii'))
 
@@ -109,6 +109,21 @@ if __name__ == '__main__':
       help =
         'Number of addresses to generate. '
         'If not specified, the first unused address will be returned.'
+  )
+
+  parser.add_argument(
+    '--security',
+      type      = int,
+      default   = 2,
+      help      = 'Security level to be used for the private key / address. Can be 1, 2 or 3',
+  )
+
+  parser.add_argument(
+    '--with-checksum',
+      action    = 'store_true',
+      default   = False,
+      dest      = 'checksum',
+      help      = 'List the address with the checksum.',
   )
 
   main(**vars(parser.parse_args(argv[1:])))
--- a/examples/address_generator.py
+++ b/examples/address_generator.py
@@ -12,12 +12,13 @@ from sys import argv
 from typing import Optional, Text
 
 from iota import __version__, Iota
+from iota.crypto.addresses import AddressGenerator
 from iota.crypto.types import Seed
 from six import binary_type, moves as compat, text_type
 
 
 def main(uri, index, count, security, checksum):
-  # type: (Text, int, Optional[int], Optional[int], Optional[bool]) -> None
+  # type: (Text, int, Optional[int], Optional[int], bool) -> None
   seed = get_seed()
 
   # Create the API instance.
@@ -114,8 +115,9 @@ if __name__ == '__main__':
   parser.add_argument(
     '--security',
       type      = int,
-      default   = 2,
-      help      = 'Security level to be used for the private key / address. Can be 1, 2 or 3',
+      default   = AddressGenerator.DEFAULT_SECURITY_LEVEL,
+      help      = 'Security level to be used for the private key / address. '
+                  'Can be 1, 2 or 3',
   )
 
   parser.add_argument(
--- a/iota/api.py
+++ b/iota/api.py
@@ -611,8 +611,9 @@ class Iota(StrictIota):
       index = 0,
       count = 1,
       security_level = AddressGenerator.DEFAULT_SECURITY_LEVEL,
+      checksum = False,
   ):
-    # type: (int, Optional[int], int) -> dict
+    # type: (int, Optional[int], int, Optional[bool]) -> dict
     """
     Generates one or more new addresses from the seed.
 
@@ -636,6 +637,10 @@ class Iota(StrictIota):
 
       This value must be between 1 and 3, inclusive.
 
+    :param checksum:
+      Specify whether to return the address with the checksum.
+      Defaults to False.
+
     :return:
       Dict with the following items::
 
@@ -651,6 +656,7 @@ class Iota(StrictIota):
       count         = count,
       index         = index,
       securityLevel = security_level,
+      checksum      = checksum,
       seed          = self.seed,
     )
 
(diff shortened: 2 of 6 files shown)


todofixthis:
Hey @plenarius thanks for your pull request!!  I will take a look this weekend!

examples/address_generator.py
```diff
@@ -16,8 +16,8 @@
 from six import binary_type, moves as compat, text_type
 
 
-def main(uri, index, count):
-  # type: (Text, int, Optional[int], bool) -> None
+def main(uri, index, count, security, checksum):
+  # type: (Text, int, Optional[int], Optional[int], Optional[bool]) -> None
```
todofixthis: `Optional[bool]` means that the value can be either a boolean or `None` ([PEP-484](https://www.python.org/dev/peps/pep-0484/#union-types)).

In this case, though, `None` is not an expected value for `checksum`, so we don't need the `Optional` part.

examples/address_generator.py
```diff
@@ -111,4 +111,19 @@ def output_seed(seed):
         'If not specified, the first unused address will be returned.'
   )
 
+  parser.add_argument(
+    '--security',
+      type      = int,
+      default   = 2,
+      help      = 'Security level to be used for the private key / address. Can be 1, 2 or 3',
+  )
```
todofixthis: This is a great idea!  Good thinking!

Minor thing:  Could we replace the `2` with `AddressGenerator.DEFAULT_SECURITY_LEVEL`?

examples/address_generator.py
```diff
@@ -111,4 +111,19 @@ def output_seed(seed):
...
+
+  parser.add_argument(
+    '--with-checksum',
+      action    = 'store_true',
+      default   = False,
+      dest      = 'checksum',
+      help      = 'List the address with the checksum.',
+  )
```
todofixthis: 👍

iota/api.py
```diff
@@ -611,8 +611,9 @@ def get_new_addresses(
       index = 0,
       count = 1,
       security_level = AddressGenerator.DEFAULT_SECURITY_LEVEL,
+      checksum = False,
   ):
-    # type: (int, Optional[int], int) -> dict
+    # type: (int, Optional[int], int, Optional[bool]) -> dict
```
todofixthis: `Optional` not needed, etc., etc.

iota/api.py
```diff
@@ -636,6 +637,10 @@ def get_new_addresses(
 
       This value must be between 1 and 3, inclusive.
 
+    :param checksum:
+      Specify whether to return the address with the checksum.
+      Defaults to False.
```
todofixthis: I feel like there's some documentation we could include a link to here, but I can't find any ¯\\\_(ツ)\_/¯

_Side note: typing the shrug dude in a Markdown editor sure is tricky!_

iota/commands/extended/get_new_addresses.py
```diff
@@ -36,25 +36,35 @@ def _execute(self, request):
     count           = request['count'] # type: Optional[int]
     index           = request['index'] # type: int
     security_level  = request['securityLevel'] # type: int
+    checksum        = request['checksum'] # type: Optional[bool]
```
todofixthis: Don't need `Optional`, yadda yadda yadda...

iota/commands/extended/get_new_addresses.py
```diff
@@ -36,25 +36,35 @@ def _execute(self, request):
...
       # transactions.
       for addy in generator.create_iterator(start=index):
-        response = FindTransactionsCommand(self.adapter)(addresses=[addy])
+        # If we're generating addresses with checksums we need to check
+        # for transactions on the address without the checksum
+        if not checksum:
+          response = FindTransactionsCommand(self.adapter)(addresses=[addy])
+        else:
+          response = FindTransactionsCommand(self.adapter)(
+            addresses=[addy][:-AddressChecksum.LEN]
```
todofixthis: Alternatively, we could use `addy.address` here.

https://github.com/iotaledger/iota.lib.py/blob/2.0.2/iota/types.py#L698-L699

iota/commands/extended/get_new_addresses.py
```diff
@@ -84,12 +94,15 @@ def __init__(self):
             | f.Max(self.MAX_SECURITY_LEVEL)
             | f.Optional(default=AddressGenerator.DEFAULT_SECURITY_LEVEL),
 
+        'checksum':  f.Type(bool) | f.Optional(default=False),
```
todofixthis: 👍 

`f.Optional` specifies a default value that will be used if the incoming value is `None`:
https://filters.readthedocs.io/en/latest/filters_list.html#miscellaneous-filters

This filter is occasionally important, because all the other filters will treat `None` as a passthrough value by default:
https://filters.readthedocs.io/en/latest/getting_started.html#much-ado-about-none

iota/commands/extended/get_new_addresses.py
```diff
@@ -84,12 +94,15 @@ def __init__(self):
...
+
         'seed': f.Required | Trytes(result_type=Seed),
       },
 
       allow_missing_keys = {
         'count',
         'index',
         'securityLevel',
+        'checksum',
```
todofixthis: Could we keep these strings in alphabetical order, pretty please?

pwease kitty

iota/crypto/addresses.py
```diff
@@ -175,7 +176,10 @@ def _generate_address(self, key_iterator):
 
     Used in the event of a cache miss.
     """
-    return self.address_from_digest(self._get_digest(key_iterator))
+    if self.checksum:
+      return self.address_from_digest(self._get_digest(key_iterator)).with_valid_checksum()
+    else:
+      return self.address_from_digest(self._get_digest(key_iterator))
```
todofixthis: All that work just for 3 extra lines of code :P

test/commands/extended/get_new_addresses_test.py
```diff
@@ -446,3 +475,20 @@ def test_get_addresses_online(self):
...
+        count         = 1,
+        index         = 0,
+        seed          = self.seed,
+        checksum      = True,
+      )
+
+    self.assertDictEqual(
+      response,
+      {'addresses': [self.addy_1_checksum]},
+    )
```
todofixthis: These tests look great, thank you!

Since `AddressGenerator._generate_address()` now also has a `checksum` argument, could you add a test to `AddressGeneratorTestCase` as well?

That way, if we ever modify `GetNewAddressesCommand` and change the way it does checksums, we will still have coverage for `AddressGenerator`'s checksum functionality.

iota/commands/extended/get_new_addresses.py
```diff
@@ -36,25 +36,35 @@ def _execute(self, request):
...
-  def _find_addresses(self, seed, index, count, security_level):
-    # type: (Seed, int, Optional[int], int) -> List[Address]
+  def _find_addresses(self, seed, index, count, security_level, checksum):
+    # type: (Seed, int, Optional[int], int, Optional[bool]) -> List[Address]
     """
     Find addresses matching the command parameters.
     """
-    # type: (Seed, int, Optional[int]) -> List[Address]
-    generator = AddressGenerator(seed, security_level)
+    # type: (Seed, int, Optional[bool]) -> List[Address]
```
todofixthis: I think this guy's a typo, actually.  The type hint goes above the docstring, not under it ([PEP-484](https://www.python.org/dev/peps/pep-0484/#suggested-syntax-for-python-2-7-and-straddling-code)).  You can remove this line entirely.

iota/commands/extended/get_new_addresses.py
plenarius: Aha, I was looking for that! I don't know why I didn't even guess it. Much cleaner.

PR #107, opened 2017-12-10
Simplify ProposedMultisigBundle

bhargavrpatel:
**Context**
The current ProposedMultisigBundle validates various conditions and if
applicable, raises exceptions. At a brief glance, the code did not look
as human friendly as it can be.

The proposed changes seek to simply the module as well as breakup the
code into logical pieces which makes it easy to reason with. As a
consequence, the code is also much more compact.

**Due Diligence**
It seems the original tests might be failing. There was a hardcoded rejection of inputs length greater than 1 but I see some tests passing ~5 params. I've validated that I did not fail any other tests, but I'd like some clarification related to this.

\---

**Questions for Gatekeepers**
1. What style guide/PEP standard are you using as it does not seem to be PEP8
2. To that end, what docstring format are we using? It seems to be inconsistent at the moment. Are we open to considering documenting as per the folks at [numpy](sphinxcontrib-napoleon.readthedocs.io/en/latest/example_numpy.html)? That docstring is autodetected by Sphinx which we can use for documentation.

\---

**Note**
\- First contribution to the project ⚡️⚡️⚡️ . I hope its not a major change. I tried to join Slack to pitch the change first. However, the invite has expired.
\- This branch was originally based off of master hence the `master-` prefix. However, I've made the change locally to have the branch based off of develop and cherry-picked the changed. Branch name had to be kept the same to avoid having this PR disappear.

### DIFF
--- a/iota/multisig/transaction.py
+++ b/iota/multisig/transaction.py
@@ -32,65 +32,49 @@ class ProposedMultisigBundle(ProposedBundle):
 
       Note: at this time, only a single multisig input is supported.
     """
+
     if self.hash:
       raise RuntimeError('Bundle is already finalized.')
 
-    count = 0
-    for addy in inputs:
-      if count > 0:
-        raise ValueError(
-          '{cls} only supports 1 input.'.format(cls=type(self).__name__),
+    if not inputs or len(inputs) != 1:
+      raise ValueError(
+        '{cls} requires input MultisigAddresses and currently only supports 1'
+        ' input.'
+        .format(cls=type(self).__name__),
+      )
+
+    # Validate all addresses
+    for address in inputs:
+      exc = None
+      security_level = address.security_level
+      exception_context = {'actual_input': address}
+
+      if not isinstance(address, MultisigAddress):
+        exc = TypeError(
+          'Incorrect input type for {cls} '
+          '(expected {expected}, actual {actual}).'.format(
+            actual    = type(address).__name__,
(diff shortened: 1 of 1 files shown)


todofixthis:
Hey @bhargavrpatel thank you for your pull request!!

👀  I will take a look at this later this week.  In the meantime, could you log an issue on the issue tracker outlining the problem?  This will give the community a chance to review and respond to it before the changes are merged.

📨  I'll get you an invite to the IOTA Slack; ping me at https://forum.iota.org/u/phx and I'll set you up.

💥 Quick thoughts about the failing tests:  The tests fail in Python 3.  I didn't look too closely, but first thought is that they fail because the code is using `map`, which returns a generator in Python 3.  Maybe try replacing it with an explicit loop.

✏️ WRT coding style, have a look at [CONTRIBUTING.rst](CONTRIBUTING.rst):

> - Please ensure that your code conforms to PEP-8.
>     - You may deviate from PEP-8 if you feel that your changes improve readability, but be aware that you may be asked to justify your decision.

📘 WRT docstrings, we should be using reStructuredText syntax.  If you find any errors or inconsistencies, feel free to propose a fix (:

bhargavrpatel:
@todofixthis Thanks!

I am confused and do not know if making an issue makes sense here. All the changes do is simplify the logic in the module. Its not a new functionality, nor does it fix any bugs. Please advice.

PR #108, opened 2017-12-11
Implementation of IsReattachableCommand

jinnerbichler:
Hey,

I've implemented the functionality to check whether a reattachment makes sense for certain addresses/transactions.

Best,
  Hannes

### DIFF
--- a/iota/commands/extended/__init__.py
+++ b/iota/commands/extended/__init__.py
@@ -22,3 +22,4 @@ from .prepare_transfer import *
 from .replay_bundle import *
 from .send_transfer import *
 from .send_trytes import *
+from .is_reattachable import *
--- a/iota/commands/extended/__init__.py
+++ b/iota/commands/extended/__init__.py
@@ -18,8 +18,8 @@ from .get_inputs import *
 from .get_latest_inclusion import *
 from .get_new_addresses import *
 from .get_transfers import *
+from .is_reattachable import *
 from .prepare_transfer import *
 from .replay_bundle import *
 from .send_transfer import *
 from .send_trytes import *
-from .is_reattachable import *
--- a/test/commands/extended/is_reattachable_test.py
+++ b/test/commands/extended/is_reattachable_test.py
@@ -59,24 +59,6 @@ class IsReattachableRequestFilterTestCase(BaseFilterTestCase):
       },
     )
 
-  # def test_pass_single_address(self):
-  #   """
-  #   The incoming request contains a single address
-  #   """
-  #   request = {
-  #     'addresses': Address(self.address_1)
-  #   }
-  #
-  #   filter_ = self._filter(request)
-  #
-  #   self.assertFilterPasses(filter_)
-  #   self.assertDictEqual(
-  #     filter_.cleaned_data,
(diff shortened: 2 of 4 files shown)


iota/commands/extended/__init__.py
```diff
@@ -22,3 +22,4 @@
 from .replay_bundle import *
 from .send_transfer import *
 from .send_trytes import *
+from .is_reattachable import *
```
todofixthis: I'd like to keep these guys in alphabetical order.  Makes it easier to tell at a glance if something is missing.

test/commands/extended/is_reattachable_test.py
```diff
@@ -0,0 +1,191 @@
...
+  def test_fail_empty(self):
+    """
+    The incoming request is empty.
+    """
+    self.assertFilterErrors(
+      {},
+      {
+        'addresses': [f.FilterMapper.CODE_MISSING_KEY],
+      },
+    )
```
todofixthis: Could we include a test for `addresses` not an array (e.g., calling `is_reattachable(Address(...))`)?

todofixthis:
Hey @jinnerbichler I feel like this guy is really close to being ready to merge.  Looks like you put a decent amount of work into it; would be awesome to integrate it into the library proper!

I've requested a few changes, but it's mostly minor stuff.  Let me know if you have any questions, or if I can provide additional info.

iota/commands/extended/__init__.py
jinnerbichler: done

test/commands/extended/is_reattachable_test.py
jinnerbichler: How is that possible with filters (https://filters.readthedocs.io/en/latest/#)?

I couldn't find a proper configuration for accepting both, an `Address` OR an array containing `Address` objects?

That's how validation done at the time beeing:
```
class IsReattachableRequestFilter(RequestFilter):
  def __init__(self):
    super(IsReattachableRequestFilter, self).__init__(
      {
        'addresses': (
          f.Required
          | f.Array
          | f.FilterRepeater(
            f.Required
            | Trytes(result_type=Address)
            | f.Unicode(encoding='ascii', normalize=False)
          )
        )
      }
    )
```

jinnerbichler:
@todofixthis Can you restart the CI? It had some problems fetching proper requirements.

test/commands/extended/is_reattachable_test.py
todofixthis: Sorry; I should have clarified — if `addresses` is not an array, the filter should reject it.

Just for the sake of coverage — if we were to remove `f.Array` from that filetr, I don't think there are any unit tests that would fail as a result.

todofixthis:
All set!  3.6 build finished successfully.  Sorry again for the ambiguous wording on my last comment.

Once that last test is in place, we'll get this guy merged for the next release!

test/commands/extended/is_reattachable_test.py
jinnerbichler: np. Added this test case

PR #109, opened 2017-12-13
Adding multisignature address doc

QuantumDamage:
I added multisig doc, and extended it with commentaries about signatures order. I also copied warning from wiki to the end of document so they should be easier to find when somebody is quickly searching for them.


todofixthis:
Thanks @QuantumDamage !!  I will take a look at this over the weekend.

PR #110, opened 2017-12-15
Added settable timeout in HTTP adapter

### LINKED ISSUE #105: Add `timeout` argument to `HttpAdapter.__init__`
Right now, `iota.adapter.HttpAdapter` uses `socket.getdefaulttimeout` to set the timeout for each HTTP request.<sup>1</sup>

However, applications really should have the flexibility to define a timeout specific to IOTA API requests (and even potentially to set different timeouts per `iota.api.Iota` instance in the same application).

Add a `timeout` argument to `HttpAdapter.__init__`.  If not specified it should default to `None`<sup>2</sup>.

When `iota.adapter.HttpAdapter._send_http_request` runs, it should use the following logic to determine the timeout value:

- If `kwargs['timeout']` is set, use that.
- Else if `self.timeout` is set, use that.
- Else use `socket.getdefaulttimeout()`.

<sup>1</sup> Technically, this can be set on a per-request basis, but it is at such a low level that regular applications won't be able to use it.

<sup>2</sup> **Important:** the default value for `timeout` should be `None`, **not** `socket.getdefaulttimeout()`.  That way, if the application calls `socket.setdefaulttimeout()` at any point after creating the instance, `HttpAdapter` doesn't cache the stale value.

jinnerbichler:
Fixes #105

### DIFF
--- a/iota/adapter/__init__.py
+++ b/iota/adapter/__init__.py
@@ -223,8 +223,8 @@ class HttpAdapter(BaseAdapter):
   in the ``headers`` kwarg.
   """
 
-  def __init__(self, uri, timeout=None, **kwargs):
-    # type: (Union[Text, SplitResult], Optional(int)) -> None
+  def __init__(self, uri, timeout=None):
+    # type: (Union[Text, SplitResult], Optional[int]) -> None
     super(HttpAdapter, self).__init__()
 
     self.timeout = timeout
--- a/iota/adapter/__init__.py
+++ b/iota/adapter/__init__.py
@@ -32,11 +32,13 @@ if PY2:
   # https://travis-ci.org/iotaledger/iota.lib.py/jobs/191974244
   __all__ = map(binary_type, __all__)
 
+
 API_VERSION = '1'
 """
 API protocol version.
 https://github.com/iotaledger/iota.lib.py/issues/84
 """
+
 
 # Custom types for type hints and docstrings.
 AdapterSpec = Union[Text, 'BaseAdapter']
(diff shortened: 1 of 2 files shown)


iota/adapter/__init__.py
```diff
@@ -223,10 +223,12 @@ class HttpAdapter(BaseAdapter):
   in the ``headers`` kwarg.
   """
 
-  def __init__(self, uri):
-    # type: (Union[Text, SplitResult]) -> None
+  def __init__(self, uri, timeout=None, **kwargs):
+    # type: (Union[Text, SplitResult], Optional(int)) -> None
```
todofixthis: Type hint s/b `Optional[int]` w/ square brackets.

Also, is `kwargs` used?

test/adapter_test.py
```diff
@@ -314,6 +315,43 @@ def test_non_object_response(self):
...
+    _, kwargs = request_mock.call_args
+    self.assertEqual(kwargs['timeout'], 88)
+
+    request_mock.request_mock()
+
+    # test with timeout at adapter creation
+    adapter = HttpAdapter('http://localhost:14265', timeout=99)
+    adapter.send_request(payload=mock_payload)
+    _, kwargs = request_mock.call_args
+    self.assertEqual(kwargs['timeout'], 99)
```
todofixthis: Could we split these into separate tests?  That will ensure that each of the test scenarios run, even if one of the assertions fails — avoiding the situation where you fix the code to make an assertion pass... only for the very next assertion in the test to fail.

jinnerbichler: done

todofixthis: Thanks!

todofixthis: Could we give each of these tests a different name, to make it easier to tell what exactly is being tested in each scenario?

Also, should lines 364-368 also be refactored into their own test?

todofixthis:
Hey @jinnerbichler I think this PR is just about ready to merge; just a couple of minor changes to the unit tests.  It would be great to get this into the next release; let me know if you have any questions, or if I can provide additional info.

Issue #101, opened 2017-11-21
Chop off checksum when using address in core commands


todofixthis:
Example:
```python
In [1]: from iota import Address

In [2]: addy = Address('WLVZXWARPSYCWJMBZJGXHUOVYBVCEKMNQDMXHCAEJZFLFLMHFYYQQSSLVYWAZWESKXZOROLU9OQFRVDEWCHKKXPWGY')

In [3]: api.get_balances([addy])
iota.adapter.BadApiResponse: 400 response from node: Invalid addresses input
```

Workaround:
```python
In [4]: api.get_balances([addy.address])
Out[4]:
{'balances': [0],
 'duration': 9,
 'milestone': Address(b'GNKYBMEKZFBYEQXINYTUGLGVSYCMRBBYDCWJ9KMJUTVMNDUVZQKGJK9RPBGTLWPYYERSLBPQBUTRZ9999'),
 'milestoneIndex': 286176}
```

I feel like PyOTA should do this automatically — there are a lot of use cases that involve generating addresses with checksums; it would get pretty unwieldy to have to remember to extract just the address when calling certain commands.

plenarius:
Gonna look into tackling this over the next week.

plenarius:
Okay, the only two core commands that take addresses as arguments are `get_balances` and `find_transactions`. My first shot at this has me massaging the addresses in a new `_prepare_requests` for both `GetBalancesCommand` and `FindTransactionsCommand` with such:

```python
def _prepare_request(self, request):
    no_checksum_addresses = []

    for addy in request['addresses']:
      no_checksum_addresses.append(Address(addy.address))

    request['addresses'] = no_checksum_addresses

    return self._apply_filter(
      value           = request,
      filter_         = self.get_request_filter(),
      failure_message = 'Request failed validation',
    )
```
Sound like an okay approach?

todofixthis:
Could we integrate this functionality directly into the filters?  That way, we don't have to worry about copy-pasting that code if we add new commands in the future (I'm aware of at least 2 that are coming in the next IRI release).

Take a look at these guys and see if it makes sense to implement a filter:
https://github.com/iotaledger/iota.lib.py/blob/develop/iota/filters.py

plenarius:
Yep I can certainly do that, looked all into that already but thought it may be overkill. Will see what I come up with.

todofixthis:
Scheduled for release: 2.0.3

PR #112, opened 2017-12-19
Initial work on #101 - Filter that automatically removes checksums for core commands

plenarius:
Not sure if you want some extended test cases in the get_balances and find_transactions test units as well.

Also wonder what you think about whether to accept bad checksummed addresses anyway or give an error as I am?

### DIFF
--- a/iota/commands/core/find_transactions.py
+++ b/iota/commands/core/find_transactions.py
@@ -7,7 +7,7 @@ from six import iteritems
 
 from iota import Address, Tag, TransactionHash
 from iota.commands import FilterCommand, RequestFilter, ResponseFilter
-from iota.filters import Trytes, AddressNoChecksum
+from iota.filters import AddressNoChecksum, Trytes
 
 __all__ = [
   'FindTransactionsCommand',
--- a/iota/commands/core/find_transactions.py
+++ b/iota/commands/core/find_transactions.py
@@ -43,7 +43,7 @@ class FindTransactionsRequestFilter(RequestFilter):
             f.Array
           | f.FilterRepeater(
                 f.Required
-              | AddressNoChecksum(result_type=Address)
+              | AddressNoChecksum()
               | f.Unicode(encoding='ascii', normalize=False)
             )
         ),
--- a/iota/commands/core/find_transactions.py
+++ b/iota/commands/core/find_transactions.py
@@ -7,7 +7,7 @@ from six import iteritems
 
 from iota import Address, Tag, TransactionHash
 from iota.commands import FilterCommand, RequestFilter, ResponseFilter
-from iota.filters import Trytes
+from iota.filters import Trytes, AddressNoChecksum
 
 __all__ = [
   'FindTransactionsCommand',
@@ -43,7 +43,7 @@ class FindTransactionsRequestFilter(RequestFilter):
             f.Array
           | f.FilterRepeater(
                 f.Required
-              | Trytes(result_type=Address)
+              | AddressNoChecksum(result_type=Address)
               | f.Unicode(encoding='ascii', normalize=False)
             )
         ),
--- a/iota/filters.py
+++ b/iota/filters.py
@@ -161,20 +161,14 @@ class AddressNoChecksum(Trytes):
   def _apply(self, value):
     super(AddressNoChecksum, self)._apply(value)
 
+    if self._has_errors:
+      return None
+
+    # Possible it's still just a TryteString
     if not isinstance(value, Address):
-      try:
-          value = Address(value)
-      except ValueError:
-          return self._invalid_value(
-            value     = value,
-            reason    = self.CODE_WRONG_FORMAT,
-          exc_info  = True,
+      value = Address(value)
 
-          template_vars = {
-              'result_type': self.result_type.__name__,
-          },
-        )
-
+    # Bail out if we have a bad checksum
     if value.checksum and not value.is_checksum_valid():
       return self._invalid_value(
         value     = value,
--- a/iota/filters.py
+++ b/iota/filters.py
@@ -147,7 +147,7 @@ class Trytes(f.BaseFilter):
 
 class AddressNoChecksum(Trytes):
   """
-  Validates a sequence as a sequence of trytes.
+  Validates a sequence as an Address then chops off the checksum if it exists
   """
 
   def _apply(self, value):
@@ -163,6 +163,7 @@ class AddressNoChecksum(Trytes):
 
     # If the incoming value already is an Address then we just make sure it has no checksum
     if isinstance(value, Address):
+      # Make sure it has a valid checksum if one exists
       if value.checksum and not value.is_checksum_valid():
         return self._invalid_value(
         value     = value,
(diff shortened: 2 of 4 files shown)


test/filters_test.py
```diff
@@ -201,3 +201,48 @@ def test_fail_wrong_type(self):
...
+    Incoming value is not an :py:class:`Address` instance.
+    """
+    self.tryte1 = (
+      b'TESTVALUE9DONTUSEINPRODUCTION99999FBFFTG'
+      b'QFWEHEL9KCAFXBJBXGE9HID9XCOHFIDABHDG9AHDR'
+    )
+    self.checksum = b'ENXYJOBP9'
+    self.address = Address(self.tryte1)
+    self.address_with_checksum = Address(self.tryte1 + self.checksum)
+    self.address_with_bad_checksum = Address(self.tryte1 + b'DEADBEEF9')
```
todofixthis: 🐮 🔫 ❗️ 
😸

test/filters_test.py
```diff
@@ -201,3 +201,48 @@ def test_fail_wrong_type(self):
...
+
+  def test_pass_checksumless_addy(self):
```
todofixthis: 😸 Not to be confused with addresses that have multiple checksums — checksummore addys :P

test/filters_test.py
```diff
@@ -201,3 +201,48 @@ def test_fail_wrong_type(self):
...
+    self.assertFilterPasses(self.address_with_checksum, self.address)
+
+  def test_fail_badchecksum_addy(self):
+    """
+    If they've got a bad checksum in their address we should probably tell
+    them so they don't wonder why something works in one place and not another
+    """
+    self.assertFilterErrors(
+      self.address_with_bad_checksum,
+      [AddressNoChecksum.ADDRESS_BAD_CHECKSUM])
```
todofixthis: 👍 This is a really cool idea.  I like it.

I feel like this feature will be received positively (grudgingly so, perhaps, but positively nonetheless).

If a developer is super confident that they know what they are doing, they can always strip off the checksum before providing it to the API method.  But they have to explicitly bypass the safety mechanism, so it forces the developer to really think about what they're doing (and whether maybe – just maybe – the invalid checksum is because of a bug somewhere else in the code that needs to be addressed).

iota/commands/core/find_transactions.py
```diff
@@ -43,7 +43,7 @@ def __init__(self):
             f.Array
           | f.FilterRepeater(
                 f.Required
-              | Trytes(result_type=Address)
+              | AddressNoChecksum(result_type=Address)
```
todofixthis: ❔ Is it possible to design `AddressNoChecksum` so that its initialiser doesn't require a `result_type` argument?  Or at least, make it optional?

I'm trying to think of a scenario where a developer would want to use something other than `Address` there, but I can't think of anything.

iota/filters.py
```diff
@@ -142,3 +143,67 @@ def _apply(self, value):
...
+    if isinstance(value, Address):
+      if value.checksum and not value.is_checksum_valid():
+        return self._invalid_value(
+        value     = value,
+        reason    = self.ADDRESS_BAD_CHECKSUM,
+        exc_info  = True,
+
+        template_vars = {
+          'result_type': self.ADDRESS_BAD_CHECKSUM,
+        },
```
todofixthis: ❔ I'm not sure that `result_type` is necessary here.  `template_vars` contains values that will be used to format the template string.  For an example, see the `Min` filter:

- https://github.com/eflglobal/filters/blob/1.3.2/filters/number.py#L223
- https://github.com/eflglobal/filters/blob/1.3.2/filters/number.py#L272-L273

In this case, there are no placeholders in the template, so `template_vars` is probably not needed.

That said, there is a `context` argument where we can attach extra values to the resulting error, to help the developer troubleshoot the error.  For example, we could do this:

```python
return self._invalid_value(
  ...
  context = {
    'expected_checksum': value.with_valid_checksum().checksum,
  }
)
```

iota/filters.py
```diff
@@ -142,3 +143,67 @@ def _apply(self, value):
...
+      ) # type: TrytesCompatible
+
+    if self._has_errors:
+      return None
+
+    # If the incoming value already is an Address then we just make sure it has no checksum
+    if isinstance(value, Address):
+      if value.checksum and not value.is_checksum_valid():
+        return self._invalid_value(
+        value     = value,
```
todofixthis: ❔ I think these lines need indentation++.

iota/filters.py
```diff
@@ -142,3 +143,67 @@ def _apply(self, value):
...
+        },
+      )
+      return Address(value.address)
+
+    # First convert to a generic TryteString, to make sure that the
+    # sequence doesn't contain any invalid characters.
+    try:
+      value = TryteString(value)
+    except ValueError:
+      return self._invalid_value(value, self.CODE_NOT_TRYTES, exc_info=True)
```
todofixthis: ❔ Since `AddressNoChecksum` (prudently) derives from `Trytes`, I think we might be able to replace a lot of this logic with a call to `super(AddressNoChecksum, self)._apply(value)`.

iota/filters.py
```diff
@@ -142,3 +143,67 @@ def _apply(self, value):
...
+    except ValueError:
+      return self._invalid_value(
+        value     = value,
+        reason    = self.CODE_WRONG_FORMAT,
+        exc_info  = True,
+
+        template_vars = {
+          'result_type': self.result_type.__name__,
+        },
+      )
```
todofixthis: ☂️ Is it possible to re-engineer this method so that the checksum checker code only appears once?

iota/filters.py
```diff
@@ -69,10 +68,12 @@ class Trytes(f.BaseFilter):
   """
   Validates a sequence as a sequence of trytes.
   """
-  CODE_NOT_TRYTES   = 'not_trytes'
-  CODE_WRONG_FORMAT = 'wrong_format'
+  ADDRESS_BAD_CHECKSUM   = 'address_bad_checksum'
```
todofixthis: ❔ Should this guy live in the `AddressNoChecksum` class instead of `Trytes`?

💡 We don't have to worry about `templates` getting overwritten; the metaclass will combine them for us automatically:
https://github.com/eflglobal/filters/blob/1.3.2/filters/base.py#L42-L48

test/filters_test.py
```diff
@@ -201,3 +201,48 @@ def test_fail_wrong_type(self):
...
+    self.address_with_bad_checksum = Address(self.tryte1 + b'DEADBEEF9')
+
+  def test_pass_checksumless_addy(self):
+    """
+    Incoming value is tryte in address form or Address object
+    """
+    self.assertFilterPasses(self.tryte1)
+    self.assertFilterPasses(self.address)
+
+  def test_pass_withchecksum_addy(self):
```
todofixthis: ❔ I think I see what you're going for here with the naming convention.  I like the idea; could we maybe use camelCase instead of smashcase in the test names?  E.g., `test_pass_withChecksum_addy`.

I find camelCase easier to read, and IDEs are better at identifying word boundaries in camelCase.

iota/filters.py
plenarius: Oh thank you for the lesson, I didn't know the metaclass combines them. Will do.

iota/filters.py
plenarius: I will have to learn a bit more with how super works and how you combine your own logic in your sub with it, but doesn't sound like it'll be too hard.

iota/commands/core/find_transactions.py
plenarius: Fair, I think that will inherently clean up a bit too.

test/filters_test.py
plenarius: Just a heads up this is ready for review again.

Issue #121, opened 2017-12-26
Add isPromotable and promoteTransaction commands


plenarius:
With `checkConsistency` now added in #113 we can add `isPromotable` and `promoteTransaction`

[isPromotable](https://github.com/iotaledger/iota.lib.js/blob/433f5d37a684abf875aaa6f1fb97082ab06218f0/lib/api/api.js#L1944)

[promoteTransaction](https://github.com/iotaledger/iota.lib.js/blob/433f5d37a684abf875aaa6f1fb97082ab06218f0/lib/api/api.js#L632)

scottbelden:
I have most of this locally that I need to push to a branch and make a PR for. I'll try to do that soon.

scottbelden:
I think this should be complete now.

PR #124, opened 2017-12-27
add is_promotable api

scottbelden:
The first part of https://github.com/iotaledger/iota.lib.py/issues/121

I don't know if this is overkill, but I wasn't sure how to handle this...

`is_promotable` could just simply be the following:
```python
  def is_promotable(self, tail):
    # type: (TransactionHash) -> bool
    """
    Determines if a tail transaction is promotable.

    :param tail:
      Transaction hash. Must be a tail transaction.
    """
    return self.check_consistency(tails=[tail])['state']
```


iota/commands/extended/is_promotable.py
```diff
@@ -0,0 +1,40 @@
...
+class IsPromotableCommand(FilterCommand):
+  """
+  Executes ``isPromotable`` extended API command.
+
+  See :py:meth:`iota.api.Iota.is_promotable` for more info.
+  """
+  command = 'isPromotable'
+
+  def get_request_filter(self):
+    return GetBundlesRequestFilter()
```
todofixthis: Should this be `IsPromotableRequestFilter`?

test/commands/extended/is_promotable_test.py
```diff
@@ -0,0 +1,166 @@
...
+  def test_fail_unexpected_parameters(self):
+    """
+    Request contains unexpected parameters.
+    """
+    self.assertFilterErrors(
+      {
+        'tail': TransactionHash(self.transaction),
+
+        # SAY "WHAT" AGAIN!
+        'what': 'augh!',
```
todofixthis: Aw, no fair!  You can't just copy-paste; you have to come up with your own joke here 😺

iota/commands/extended/is_promotable.py
scottbelden: Yes. Fixed. I'm sure the CI would have eventually caught that with a `NameError` but you beat it to it

todofixthis:
I think it's fine that `IsPromotableCommand` is basically just a wrapper for `checkConsistency`.

It follows the existing convention, and it gives us room to grow if the IOTA Foundation adds more logic to this process.

When designing the `Iota` class, I tried to follow the [Command Pattern](https://en.wikipedia.org/wiki/Command_pattern) here; depending on the use case, developers can forego the `Iota` class entirely and interact exclusively with individual command classes.

This should make it easier for developers to create their own extended commands (e.g., if a developer wants to use `isPromotable` in a custom command, they don't have to have an `Iota` instance, and they don't have to copy-paste the logic from `is_promotable()`).

scottbelden:
Makes sense. The only other thing I wasn't sure about is if the `tail` argument should be a single `Transaction` or a variable amount. It feels to me like it should be `def is_promotable(self, *tails)` since `check_consistency` can take in a list of `tails` but I just copied what the JS library did.

todofixthis:
Aye.  Nobody can fault you for mirroring the JS lib (:

In the original design for PyOTA, I took a few liberties with the API commands (e.g., snake_case instead of camelCase, `get_new_addresses` instead of `get_new_address`, etc.).  But, for the most part, I tried to stick as closely as possible to the JS lib, so that the interface is consistent.

scottbelden:
> every command returns a dict*

This API would break that convention as this just returns a `bool`. I think returning a `dict` for `is_promotable` would be awkward.

todofixthis:
Oh wow, yeah, good point; I totally missed that!

Hm...  if `is_promotable` returns a dict, then devs might as well just call `check_consistency`.

Three-ish ideas:

- `IsPromotableCommand` returns a dict, on the off chance that the IOTA core devs later introduce more complexity (I don't think this is likely to happen though).
- We set a precedent that API commands don't have to return dicts.
- We scrap `IsPromotableCommand` and move the logic directly into `Iota.is_promotable`.  This sets a different type of precedent, but we classify `is_promotable` as a convenience function rather than an API command.
- Something else?

What do you think?

scottbelden:
I don't think returning a dict is a good idea. If that were required I would say to not add this and just call `check_consistency` directly (as it would be almost identical).

I think the third bullet is reasonable since `is_promotable` isn't an API but just a helper. However, I can see that this would cause a problem if there ever were to be some official `is_promotable` API. Therefore it might be necessary to make this an obvious separation. For example:

```python
api = Iota()
api.helpers.is_promotable(transaction)
```

However, I'm okay with punting on this for the moment and waiting to make a decision until there is another more useful helper (as this only adds a small convenience and can be easily added again). 

Will [promoteTransaction](https://github.com/iotaledger/iota.lib.js/blob/433f5d37a684abf875aaa6f1fb97082ab06218f0/lib/api/api.js#L632) cause the same debate? Would it return `{'transactions': [transaction1, ...]}` or `[transaction1, ...]` (or something else)?

todofixthis:
The `helpers` idea is an interesting one.  It would also give us a home for the functions in `iota.commands.extended.utils`.

My preference is for `promote_transaction` to return a dict (`{'transactions': [...]}`).  The IRI API has a long and distinguished history of changing without warning; if we leave ourselves a little space to make changes in a backwards-compatible way, I think we (and our users!) will be a lot happier in the long run.

scottbelden:
I'll play around with adding the `helpers` and see how that looks and feels.

scottbelden:
@todofixthis I added a new commit that re-works this PR to use the `helpers` idea. Let me know what you think and then I'll clean it up and squash it down (or remove the commit and iterate again on the other way).

todofixthis:
Awesome!  I'll check it out later this week!

PR #127, opened 2018-01-02
add promote_transaction_api

scottbelden:
Resolves part two of https://github.com/iotaledger/iota.lib.py/issues/121

This is slightly different from the JS library in that the JS library takes in the `transfer` as one of the arguments. I thought this was cleaner, but I can change it back to how the JS library works if desired.

### DIFF
--- a/iota/commands/core/get_transactions_to_approve.py
+++ b/iota/commands/core/get_transactions_to_approve.py
@@ -32,7 +32,25 @@ class GetTransactionsToApproveRequestFilter(RequestFilter):
   def __init__(self):
     super(GetTransactionsToApproveRequestFilter, self).__init__({
       'depth': f.Required | f.Type(int) | f.Min(1),
+
+      'reference': Trytes(result_type=TransactionHash),
+    },
+
+    allow_missing_keys = {
+      'reference',
     })
+
+  def _apply(self, value):
+    value = super(GetTransactionsToApproveRequestFilter, self)._apply(value) # type: dict
+
+    if self._has_errors:
+      return value
+
+    # Remove reference if null.
+    if value['reference'] is None:
+      del value['reference']
+
+    return value
 
 
 class GetTransactionsToApproveResponseFilter(ResponseFilter):
--- a/iota/commands/extended/promote_transaction.py
+++ b/iota/commands/extended/promote_transaction.py
@@ -1,0 +1,68 @@
+# coding=utf-8
+from __future__ import absolute_import, division, print_function, \
+  unicode_literals
+
+import filters as f
+from iota import (
+  Bundle, TransactionHash, Address, ProposedTransaction, BadApiResponse,
+)
+from iota.commands import FilterCommand, RequestFilter
+from iota.commands.core.check_consistency import CheckConsistencyCommand
+from iota.commands.extended.send_transfer import SendTransferCommand
+from iota.filters import Trytes
+
+__all__ = [
+  'PromoteTransactionCommand',
+]
+
+
+class PromoteTransactionCommand(FilterCommand):
+  """
+  Executes ``promoteTransaction`` extended API command.
+
+  See :py:meth:`iota.api.Iota.promote_transaction` for more information.
+  """
+  command = 'promoteTransaction'
+
+  def get_request_filter(self):
+    return PromoteTransactionRequestFilter()
+
+  def get_response_filter(self):
+    pass
+
+  def _execute(self, request):
+    depth                 = request['depth'] # type: int
+    min_weight_magnitude  = request['minWeightMagnitude'] # type: int
+    transaction           = request['transaction'] # type: TransactionHash
+
+    cc_response = CheckConsistencyCommand(self.adapter)(tails=[transaction])
+    if cc_response['state'] is False:
+      raise BadApiResponse(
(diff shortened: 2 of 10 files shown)


todofixthis:
Thanks Scott!  I will take a look later this week!

iota/commands/extended/promote_transaction.py
```diff
@@ -0,0 +1,68 @@
...
+
+    cc_response = CheckConsistencyCommand(self.adapter)(tails=[transaction])
+    if cc_response['state'] is False:
+      raise BadApiResponse(
+        'Transaction {transaction} is not promotable. '
+        'You should reattach first.'.format(transaction=transaction)
+      )
+
+    spam_transfer = ProposedTransaction(
+      address=Address(b'9' * 81),
```
todofixthis: 💡 `Address` will automatically pad the value to 81 trytes.  No magic number necessary (:

test/commands/extended/promote_transaction_test.py
```diff
@@ -0,0 +1,369 @@
...
+  def test_happy_path(self):
+    """
+    Successfully promoting a bundle.
+    """
+
+    self.adapter.seed_response('checkConsistency', {
+      'state': True,
+    })
+
+    result_bundle = Bundle.from_tryte_strings(self.hash2)
```
todofixthis: ❔ Is this correct?  I think `Bundle.from_tryte_strings` expects an array of `TransactionTrytes` objects.

test/commands/extended/send_transfer_test.py
```diff
@@ -163,6 +167,7 @@ def test_pass_optional_parameters_omitted(self):
       {
         'changeAddress':      None,
         'inputs':             None,
+        'reference':          None,
```
todofixthis: ❔ Good thinking!  Could we also include unit tests for invalid values for `reference` (i.e., not TrytesCompatible)?

iota/commands/core/get_transactions_to_approve.py
```diff
@@ -32,8 +32,26 @@ class GetTransactionsToApproveRequestFilter(RequestFilter):
...
+    value = super(GetTransactionsToApproveRequestFilter, self)._apply(value) # type: dict
+
+    if self._has_errors:
+      return value
+
+    # Remove reference if null.
+    if value['reference'] is None:
+      del value['reference']
+
+    return value
```
todofixthis: ❔ Could we add a couple of unit tests for `GetTransactionsToApproveRequestFilter`, related to `reference`?

- Not TrytesCompatible.
- Included in resulting dict if not null.

scottbelden:
I rebased and believe I resolved the outstanding comments. Let me know if there's anything I missed or that should still be updated.

iota/commands/extended/promote_transaction.py
```diff
@@ -0,0 +1,68 @@
...
+
+  def get_response_filter(self):
+    pass
+
+  def _execute(self, request):
+    depth                 = request['depth'] # type: int
+    min_weight_magnitude  = request['minWeightMagnitude'] # type: int
+    transaction           = request['transaction'] # type: TransactionHash
+
+    cc_response = CheckConsistencyCommand(self.adapter)(tails=[transaction])
```
scottbelden: I wanted to use the new `is_promotable` helper here, but we don't have a reference to the `Iota` class. I don't think this is a problem as it doesn't cause much duplication, but if we have more helper functions that we want to use from these commands in the future, it might need some thought on how to achieve that.

todofixthis: Hm... interesting.

What do you think about this?

```python
class Iota:
  @property
  def helpers(self):
    return Helpers(self.adapter)
```

Then `PromoteTransactionCommand` can do `Helpers(self.adapter).is_promotable(...)`.

scottbelden: Does that mean that the current [`is_promotable`](https://github.com/iotaledger/iota.lib.py/blob/develop/iota/commands/extended/helpers.py#L10) should be changed from:

```python
  def is_promotable(self, tail):
    return self.api.check_consistency(tails=[tail])['state']
```

To:
```python
  from iota.commands.core.check_consistency import CheckConsistencyCommand

  def is_promotable(self, tail):
    return CheckConsistencyCommand(self.adapter)(tails=[tail])['state']
```

I don't see a problem with that.

todofixthis: Aye, I think that would... oh.  Hm, we might end up with circular references that way.

Eh, I guess we could import the commands locally in the helper methods.  The alternative, I think, would be to create helper classes, similar to what we're doing for commands.  That feels a bit cathedral-ish to me, though.

scottbelden: Yeah, I can foresee circular dependencies. I'm okay with keeping it as is (not using the helper call) for the moment until a later date.

scottbelden:
@todofixthis Would you be okay with merging this and creating an issue to later resolve the best way to use the helper functions? I'm not sure what's best at the moment but I wouldn't want to hold this up while trying to think of something.

todofixthis:
Hey Scott.  Sure, that sounds fine with me!  I'll go ahead and merge these changes now.

PR #140, opened 2018-01-15
Clarify Address usage in send example

maeck70:
Improved the script to explain the recipient address a little better.
Changed the address concat to a reference to a specific kind of Iota address.  
This is loosely inspired by the JAVA sen transaction test case.

### DIFF
--- a/examples/send_transfer.py
+++ b/examples/send_transfer.py
@@ -4,6 +4,8 @@ Example script that shows how to use PyOTA to send a transfer to an address.
 """
 from iota import *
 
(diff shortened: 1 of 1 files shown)


maeck70:
Example improvement due to overlooking the fact that the recipient address was shown as a two line (concat) address. Included a different way to show the fact that the address is security level 2 with validation.

maeck70:
Reopened due to incorrect close

todofixthis:
Hey @maeck70 thank you for your pull request!  I will take a look later this week.

PR #137, opened 2018-01-12
Basic authentication for HttpAdapter

HerrMuellerluedenscheid:
Addresses issue #129

Issue #129, opened 2018-01-05
Add authentication headers to HTTP requests when IRI node is running with `--remote-auth`


HerrMuellerluedenscheid:
Hi again, 

I saw the same question unanswered [on reddit ](https://www.reddit.com/r/Iota/comments/6yalf2/how_does_remoteauth_work/), so I thought I approach you directly. How can I connect to my full node when [`--remote-auth`](https://github.com/iotaledger/iri/#command-line-options) is active? I searched through the code but couldn't get the point. Is it not yet implemented or 'hidden' in a wrapping call? If it is not implemented and I wanted to give it a shot, where would I have to start digging into the code? My first naive guess would be to extend the `BaseCommand`.

Greetings from Potsdam

Marius
 

PS.: I asked on slack as well but didn't get a response, yet. And also others might sooner or later have the same question.

HerrMuellerluedenscheid:
Thanks to Nuriel Shem-Tov on slack I was able to use remote authentication by modifying the http headers of the http adapter. Here are the essentials:

```
from base64 import b64encode

adapter = HttpAdapter(
  uri = u'http://localhost:14700',
)

user_pass = b64encode(b'username:PASSWORD').decode('ascii')
adapter.DEFAULT_HEADERS[u'Authorization'] = u'Basic %s' % user_pass

api = Iota(
    adapter=adapter,
    seed = b'DMSWNXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX')
```

I'm not sure though if there is a more elegant way. Modifying _default_ header attributes seems a little workaround-ish. Should there be an 'extra_header' argument to the `BaseAdapter`?

Apart from that, the authentication can also be done right in the url as e.g. [here](https://stackoverflow.com/questions/2716990/http-basic-authentication-credentials-passed-in-url-and-encryption). But I found it a little more elegant to do it the first way.

todofixthis:
Hey @HerrMuellerluedenscheid thanks for reporting this!  I think this is a shortcoming in PyOTA — we should definitely support this out-of-the-box, no hacking necessary (:

todofixthis:
Hey @HerrMuellerluedenscheid out of curiosity, does the adapter work if username/password are included in the node URI?

E.g.:

```python
api = Iota("http://username:password@localhost:14265")
```

todofixthis:
[...]

PR #142, opened 2018-01-16
HttpAdapter docs

HerrMuellerluedenscheid:
Sane version of #137 addressing issue #129

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -70,7 +70,7 @@ Documentation
 =============
 PyOTA's documentation is available on `ReadTheDocs`_.
 
-If you are :ref:`installing from source <readme-installing-from-source>`, you
+If you are [installing from source](#installing-from-source), you
 can also build the documentation locally:
 
 #. Install extra dependencies (you only have to do this once)::
--- a/README.rst
+++ b/README.rst
@@ -70,7 +70,7 @@ Documentation
 =============
 PyOTA's documentation is available on `ReadTheDocs`_.
 
-If you are [installing from source](#installing-from-source), you
+If you are :ref:`installing from source <readme-installing-from-source>`, you
 can also build the documentation locally:
 
 #. Install extra dependencies (you only have to do this once)::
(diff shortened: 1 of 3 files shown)


README.rst
```diff
@@ -70,7 +70,7 @@ Documentation
 =============
 PyOTA's documentation is available on `ReadTheDocs`_.
 
-If you are :ref:`installing from source <readme-installing-from-source>`, you
+If you are [installing from source](#installing-from-source), you
```
todofixthis: The previous syntax is actually correct for `.rst` files.  I'll file a request with GitHub support to figure out why it's not rendering correctly.

docs/adapters.rst
```diff
@@ -44,6 +44,14 @@ HttpAdapter
     api = Iota('https://service.iotasupport.com:14265')
     api = Iota(HttpAdapter('https://service.iotasupport.com:14265'))
 
+    # Use HTTPS with basic authentication and 60 seconds timeout:
+    api = Iota('https://USERNAME:MYIRIPASSWORD@service.iotasupport.com:14265',
+               timeout=60)
```
todofixthis: ❔ I'm not sure that this works — I believe the `timeout` argument only works with an `HttpAdapter` instance.

docs/adapters.rst
```diff
@@ -44,6 +44,14 @@ HttpAdapter
...
+    api = Iota(
+        HttpAdapter(
+            'https://USERNAME:MYIRIPASSWORD@service.iotasupport.com:14265',
+            timeout=60))
```
todofixthis: 👍 Looks great!

docs/adapters.rst
HerrMuellerluedenscheid: Yes, that was a bit sloppy, copy-paste...

README.rst
todofixthis: Hey Marius.  I heard back from GitHub support.

Turns out that `:ref:` is not actually part of the reStructuredText spec; it's a custom feature provided by Sphinx.

I made the following changes:
- Removed the link.
- Added a metadata check to `.travis.yml` (which caught the `:ref:` right awy!).

Thanks for drawing attention to this issue!

PR #143, opened 2018-01-18
Fix references to "standard" API in the documentation

### LINKED ISSUE #120: Fix references to "standard" API in the documentation
The documentation contains several references to the "Standard API".  This is incorrect; it should say "Core API" instead.

Related: #118

rpitonak:
Fixes #120

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -1,5 +1,5 @@
 Core API
-============
+========
 
 The Core API includes all of the core API calls that are made
 available by the current `IOTA Reference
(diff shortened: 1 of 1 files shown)


docs/api.rst
```diff
@@ -1,14 +1,14 @@
-Standard API
+Core API
 ============
```
todofixthis: ❔ Could we adjust the `=` signs under this header so that the two lines have the same length?  It's not required by reStructuredText, but it matches the conventions of the rest of the documentation.

rpitonak:
Thanks for the review, should be fixed now.

PR #146, opened 2018-01-21
PEP8 (WIP)

HerrMuellerluedenscheid:
The PEP8 adoption could be done in one go and bundled in one pull request for all files. But it seems like it could cause quite some merge conflicts if it's not done in shortest time.

So, this is sort of a basis for discussion if the proposed style is fine (especially using 4 vs 2 spaces) for everybody. `flake8` finishes errorless.

### DIFF
--- a/iota/__init__.py
+++ b/iota/__init__.py
@@ -29,17 +29,17 @@ STANDARD_UNITS = {
 
 
 # Activate codecs.
-from .codecs import *
+from .codecs import *  # noqa
 
 # Make some imports accessible from the top level of the package.
 # Note that order is important, to prevent circular imports.
-from .types import *
-from .transaction import *
-from .adapter import *
-from .api import *
-from .trits import *
+from .types import *  # noqa
+from .transaction import *  # noqa
+from .adapter import *  # noqa
+from .api import *  # noqa
+from .trits import *  # noqa
 
 # :see: http://stackoverflow.com/a/2073599/
-from pkg_resources import require
+from pkg_resources import require  # noqa
 __version__ = require('PyOTA')[0].version
 del require
--- a/iota/api.py
+++ b/iota/api.py
@@ -1,6 +1,6 @@
 # coding=utf-8
 from __future__ import absolute_import, division, print_function, \
-  unicode_literals
+    unicode_literals
 
 from typing import Dict, Iterable, Optional, Text  # noqa
 
@@ -16,9 +16,9 @@ from iota.crypto.types import Seed
 from six import with_metaclass
 
 __all__ = [
-  'InvalidCommand',
-  'Iota',
-  'StrictIota',
+    'InvalidCommand',
+    'Iota',
+    'StrictIota',
 ]
 
 
--- a/iota/api.py
+++ b/iota/api.py
@@ -2,13 +2,14 @@
 from __future__ import absolute_import, division, print_function, \
   unicode_literals
 
-from typing import Dict, Iterable, Optional, Text
+from typing import Dict, Iterable, Optional, Text  # noqa
 
-from iota import AdapterSpec, Address, ProposedTransaction, Tag, \
-  TransactionHash, TransactionTrytes, TryteString, TrytesCompatible
-from iota.adapter import BaseAdapter, resolve_adapter
-from iota.commands import BaseCommand, CustomCommand, core, \
-  discover_commands, extended
+from iota import AdapterSpec, Address, ProposedTransaction, Tag  # noqa
+from iota import TransactionHash, TransactionTrytes, TryteString    # noqa
+from iota import TrytesCompatible   # noqa
+from iota.adapter import BaseAdapter, resolve_adapter  # noqa
+from iota.commands import BaseCommand, CustomCommand, core  # noqa
+from iota.commands import discover_commands, extended
 from iota.commands.extended.helpers import Helpers
 from iota.crypto.addresses import AddressGenerator
 from iota.crypto.types import Seed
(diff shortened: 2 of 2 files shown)


iota/__init__.py
```diff
@@ -29,17 +29,17 @@
...
-from .trits import *
+from .types import *  # noqa
+from .transaction import *  # noqa
+from .adapter import *  # noqa
+from .api import *  # noqa
+from .trits import *  # noqa
 
 # :see: http://stackoverflow.com/a/2073599/
-from pkg_resources import require
+from pkg_resources import require  # noqa
```
todofixthis: ❔ I could see why the wildcard imports would need a `# noqa`, but does this import need one as well?

iota/api.py
```diff
@@ -2,13 +2,14 @@
...
-from iota.adapter import BaseAdapter, resolve_adapter
-from iota.commands import BaseCommand, CustomCommand, core, \
-  discover_commands, extended
+from typing import Dict, Iterable, Optional, Text  # noqa
+
+from iota import AdapterSpec, Address, ProposedTransaction, Tag  # noqa
+from iota import TransactionHash, TransactionTrytes, TryteString    # noqa
+from iota import TrytesCompatible   # noqa
+from iota.adapter import BaseAdapter, resolve_adapter  # noqa
+from iota.commands import BaseCommand, CustomCommand, core  # noqa
```
todofixthis: ❔ Do these imports need `# noqa`?

iota/api.py
```diff
@@ -22,948 +23,948 @@
...
+        return core.AddNeighborsCommand(self.adapter)(uris=uris)
+
+    def attach_to_tangle(
+        self,
+        trunk_transaction,
+        branch_transaction,
+        trytes,
+        min_weight_magnitude=None
+    ):
+        # type: (TransactionHash, TransactionHash, Iterable[TryteString], int) -> dict  # noqa
```
todofixthis: ❔ Does this guy need a `# noqa`?

iota/api.py
```diff
@@ -22,948 +23,948 @@
...
+                 a list, even if only one bundle was found.
+             }
+
+        :raise:
+          - :py:class:`iota.adapter.BadApiResponse` if any of the
+            bundles fails validation.
+
+        References:
+          - https://github.com/iotaledger/wiki/blob/master/api-proposal.md#getbundle
+        """  # noqa
```
todofixthis: ❔ Do these need `# noqa`?

iota/api.py
```diff
@@ -22,948 +23,948 @@
...
-    return core.StoreTransactionsCommand(self.adapter)(trytes=trytes)
+      - https://iota.readme.io/docs/getting-started
+    """
+    commands = discover_commands('iota.commands.core')
+
+    def __init__(self, adapter, testnet=False):
+        # type: (AdapterSpec, bool) -> None
+        """
+        :param adapter:
+          URI string or BaseAdapter instance.
```
todofixthis: ❔ To be consistent, should we also use 4-space indents inside docstrings?

iota/__init__.py
HerrMuellerluedenscheid: `require` is not used. As I just read this is a warning of flake, not a PEP8 violation. But, if we include flake8 into travisCI this would be nice to get rid off. One way of handling this would be to explicitly export imported classes via [`__all__`](https://stackoverflow.com/questions/31079047/python-pep8-class-in-init-imported-but-not-used)

iota/api.py
HerrMuellerluedenscheid: According to flake8, yes. The line has more than 80 character. I was not sure if these lines are somewhat interpreted by sphinx so I didn't break that line into two shorter ones. Would that be an option instead?

iota/api.py
HerrMuellerluedenscheid: Flake8 is really picky, I have to admint. line 535 has more than 80 chars... I think the picky-ness of flake can be tuned. I'll checkout if doc strings can be ignored, if `# noqa` is undesireable.

iota/__init__.py
todofixthis: Gotcha.  Hm, that's odd.  It does get used, but then it gets removed via `del` later on (I don't want it included in the `iota` namespace).

Eh, maybe the `# noqa` here is fine.

iota/api.py
todofixthis: Gotcha.  Those are type hints, used by some IDEs (e.g., PyCharm) for autocompletion, etc.

It is possible to address the line length issue here, by doing this:

```python
def attach_to_tangle(
    self,
    trunk_transaction,  # type: TransactionHash
    branch_transaction,  # type: TransactionHash
    trytes,  # type: Iterable[TryteString]
    min_weight_magnitude=None  # type: Optional[int]
):
+        # type: (...) -> dict
```

See [PEP-484](https://www.python.org/dev/peps/pep-0484/#suggested-syntax-for-python-2-7-and-straddling-code) for more info.

iota/api.py
todofixthis: Hm.  I guess that's fine.   think it's useful to check line length on docstrings in general; if flake8 doesn't want to make exceptions for URLs, then we can live with this.

Anyway, the only alternatives I can think of are pretty awful, comparatively (split the URL into multiple lines, use URL shortener).

iota/api.py
HerrMuellerluedenscheid: I don't know. Sphinx doesn't respect formatting here anyway, right? Ah, let's make it 4 ok?

iota/api.py
HerrMuellerluedenscheid: That looks neat! :+1:

iota/api.py
HerrMuellerluedenscheid: Yeah, I also think URLs should be kept the way they are.
I'm going to update this PR asap

HerrMuellerluedenscheid:
Going to close this PR in favor of #147

Issue #145, opened 2018-01-21
Obey PEP8


miili:
Let's make the code confom with PEP8!

todofixthis:
Hey guys.  Thanks for suggesting this!

Here's what I put in [`CONTRIBUTING.rst`](https://github.com/iotaledger/iota.lib.py/blob/2.0.3/CONTRIBUTING.rst#requirements-for-pull-requests):

> - Please ensure that your code conforms to PEP-8.
>    - You may deviate from PEP-8 if you feel that your changes improve readability, but be aware that you may be asked to justify your decision.

If you'd like to take a firmer stance on PEP-8, I'm fine with that.  Maybe we can add flake8 to the Travis CI build, so that PEP-8 compliance becomes part of the acceptance criteria for new pull requests.

thalesclpr:
Well, since  I have some free type and star citizen will not launch this year, I'm open to contribute by inspecting the code and refactor it if necessary.

HerrMuellerluedenscheid:
I did a quick `flake8` (#146). Any comments about the employed style?

thalesclpr:
So, I did a quick Insect before starting working on this and [I noticed some violations](https://image.prntscr.com/image/835V3qpsTRyWQaDaygmiiA.png)

I'm going to pick this up as soon as possible.

todofixthis:
I would be hard-pressed to argue against adopting standards in an open-source project (:

The 2.0.4 release will be delayed a bit while I figure out how to update the C extension, so that should give a decent-sized window to get these changes into the next release.

todofixthis:
It would also be super cool if we could integrate flake8 into the build process ([random example](https://github.com/owocki/pytrader/commit/183d28033b56d9eec715122f9118eb3f17b2d634#diff-354f30a63fb0907d4ad57269548329e3R21)).

todofixthis:
We've made good progress on this.  All of the package files are now PEP-8-ified.  Next up is to reformat the tests (wanted to do this separately, to make it less likely that a regression would sneak in), and then I think that's it 😺

PR #147, opened 2018-01-24
Pep8

HerrMuellerluedenscheid:
Continuing #146

### DIFF
--- a/iota/api.py
+++ b/iota/api.py
@@ -1,6 +1,6 @@
 # coding=utf-8
 from __future__ import absolute_import, division, print_function, \
-  unicode_literals
+    unicode_literals
 
 from typing import Dict, Iterable, Optional, Text  # noqa
 
@@ -16,9 +16,9 @@ from iota.crypto.types import Seed
 from six import with_metaclass
 
 __all__ = [
-  'InvalidCommand',
-  'Iota',
-  'StrictIota',
+    'InvalidCommand',
+    'Iota',
+    'StrictIota',
 ]
 
 
--- a/iota/api.py
+++ b/iota/api.py
@@ -2,14 +2,8 @@
 from __future__ import absolute_import, division, print_function, \
     unicode_literals
 
-from typing import Dict, Iterable, Optional, Text  # noqa
-
-from iota import AdapterSpec, Address, ProposedTransaction, Tag  # noqa
-from iota import TransactionHash, TransactionTrytes, TryteString    # noqa
-from iota import TrytesCompatible   # noqa
-from iota.adapter import BaseAdapter, resolve_adapter  # noqa
-from iota.commands import BaseCommand, CustomCommand, core  # noqa
-from iota.commands import discover_commands, extended
+from iota.adapter import BaseAdapter, resolve_adapter
+from iota.commands import CustomCommand, core, discover_commands, extended
 from iota.commands.extended.helpers import Helpers
 from iota.crypto.addresses import AddressGenerator
 from iota.crypto.types import Seed
--- a/iota/api.py
+++ b/iota/api.py
@@ -2,13 +2,14 @@
 from __future__ import absolute_import, division, print_function, \
   unicode_literals
 
-from typing import Dict, Iterable, Optional, Text
+from typing import Dict, Iterable, Optional, Text  # noqa
 
-from iota import AdapterSpec, Address, ProposedTransaction, Tag, \
-  TransactionHash, TransactionTrytes, TryteString, TrytesCompatible
-from iota.adapter import BaseAdapter, resolve_adapter
-from iota.commands import BaseCommand, CustomCommand, core, \
-  discover_commands, extended
+from iota import AdapterSpec, Address, ProposedTransaction, Tag  # noqa
+from iota import TransactionHash, TransactionTrytes, TryteString    # noqa
+from iota import TrytesCompatible   # noqa
+from iota.adapter import BaseAdapter, resolve_adapter  # noqa
+from iota.commands import BaseCommand, CustomCommand, core  # noqa
+from iota.commands import discover_commands, extended
 from iota.commands.extended.helpers import Helpers
 from iota.crypto.addresses import AddressGenerator
 from iota.crypto.types import Seed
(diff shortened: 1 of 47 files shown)


iota/api.py
```diff
@@ -1,969 +1,968 @@
 # coding=utf-8
 from __future__ import absolute_import, division, print_function, \
-  unicode_literals
+    unicode_literals
 
-from typing import Dict, Iterable, Optional, Text
```
HerrMuellerluedenscheid: I think we have to add these again as they provide extra functionality for some smart IDEs.

miili: Or we leverage `typing` all the way through? But honestly I am not that keen on this. With a proper documentation one should know what he is doing.

https://docs.python.org/3/library/typing.html

todofixthis: Type hints are *amazing* 😺

IDEs use them for autocompletion and inspections — they're a great way to boost developer efficiency.

I agree that documentation is important, but they serve different purposes.

miili: Hi @todofixthis,
which IDE are you using?
Hope this syntactic excess of ours is not too intrusive. Once this is finished and in `dev`, we tackle the C `pearl_diver`.

todofixthis: I use PyCharm, but this is actually a standard practice; see [PEP-484](https://www.python.org/dev/peps/pep-0484/).

The syntax is a bit verbose, because we are supporting Python 2 and Python 3.  If we were only supporting Python 3, we could integrate the type hints directly into the method signature.

scottbelden:
I personally think 4 spaces is better, but making the change now is going to completely destroy the history. Also, because the change is being made with other changes it's extremely hard to see what all is being changed. If the 4 space change is going to be applied, it might be cleaner to do that specific change in it's own PR for readability sake.

rpitonak:
Hey, in #151 I changed iota/transaction/creation.py to conforms with PEP8. Can you take a look at it and check if everything is in order, so we can merge? Thank you.
@HerrMuellerluedenscheid @miili

miili:
@todofixthis which PR are we pursuing, this one or #148? We haven't put any more work in, since it seemed obsolete. Yet I disagree with 120 characters per line as allowed by #148.

todofixthis:
Hey guys.  Is the PR at a point where we could merge it, even if it's not "complete" yet?  Maybe we can merge what we've got so far, and then take a progressive approach with the remaining files.

miili:
Yessir! We can merge what is done so far.
We will continue with leftovers on this PR, if you leave if open.

todofixthis:
Okiedoke!  I'll do a quick once-over to make sure everything looks OK, and if all's well, I'll go ahead and merge it.

iota/commands/extended/send_trytes.py
```diff
@@ -1,85 +1,87 @@
 # coding=utf-8
 from __future__ import absolute_import, division, print_function, \
-  unicode_literals
+    unicode_literals
 
-from typing import List
+from typing import List  # noqa
 
-import filters as f
-from iota import TransactionTrytes, TryteString, TransactionHash
```
todofixthis: Could we put this back in, so that type hints work correctly?

iota/commands/extended/utils.py
```diff
@@ -1,19 +1,16 @@
...
 from __future__ import absolute_import, division, print_function, \
-  unicode_literals
+    unicode_literals
 
-from typing import Generator, Iterable, List, Tuple
+from typing import Generator, Iterable, List, Tuple  # noqa
 
-from iota import Address, Bundle, Transaction, \
-  TransactionHash
-from iota.adapter import BaseAdapter
```
todofixthis: Could we put this back in, so that type hints work correctly?

iota/commands/extended/utils.py
```diff
@@ -27,111 +24,106 @@ def find_transaction_objects(adapter, **kwargs):
...
 
     return []
 
 
 def iter_used_addresses(adapter, seed, start):
-  # type: (BaseAdapter, Seed, int) -> Generator[Tuple[Address, List[TransactionHash]]]
-  """
-  Scans the Tangle for used addresses.
+    # type: (BaseAdapter, Seed, int)
+    #    -> Generator[Tuple[Address, List[TransactionHash]]]
```
todofixthis: This does not work, unfortunately; the type hint has to all be on the same line.

If the type hint is too long, then it can be integrated into the function signature, like this:

```python
def iter_used_addresses(
        adapter,   # type: BaseAdapter
        seed,  # type: Seed
        start,  # type: int
):
    # type: (...) -> Generator[Tuple[Address, List[TransactionHash]], None, None]
```

(I also got the `Generator` type hint wrong; it actually requires 3 values, not 1 — the other two indicate what values can be passed in, and what value is returned; see [PEP-484](https://www.python.org/dev/peps/pep-0484/#annotating-generator-functions-and-coroutines) for details)

iota/filters.py
```diff
@@ -1,189 +1,189 @@
 # coding=utf-8
 from __future__ import absolute_import, division, print_function, \
-  unicode_literals
+    unicode_literals
 
-from typing import Text
```
todofixthis: Could we put this back in, so that type hints work correctly?

iota/json.py
```diff
@@ -1,67 +1,65 @@
 # coding=utf-8
 from __future__ import absolute_import, division, print_function, \
-  unicode_literals
+    unicode_literals
 
 from abc import ABCMeta, abstractmethod as abstract_method
 from json.encoder import JSONEncoder as BaseJsonEncoder
-from typing import Iterable, Mapping, Text
```
todofixthis: Could we put this back in, so that type hints work correctly?

iota/json.py
```diff
@@ -1,67 +1,65 @@
...
-          p.text('**')
-        elif isinstance(prepared, Iterable):
-          p.text('*')
+        References:
+            - http://ipython.readthedocs.io/en/stable/api/generated/IPython.lib.pretty.html
+            - :py:meth:`IPython.lib.pretty.RepresentationPrinter.pretty`
+            - :py:func:`pprint._safe_repr`
+        """  # noqa
+        # type: (JsonSerializable, bool) -> Text
+        cls_name = type(self).__name__
```
todofixthis: Why the rename?

iota/trits.py
```diff
@@ -7,122 +7,120 @@
 """
 
 from __future__ import absolute_import, division, print_function, \
-  unicode_literals
-
-from typing import Iterable, List, Optional, Sequence, Tuple
```
todofixthis: Could we put this back in, so that type hints work correctly?

iota/types.py
```diff
@@ -1,17 +1,16 @@
...
 from __future__ import absolute_import, division, print_function, \
-  unicode_literals
+    unicode_literals
 
 from codecs import decode, encode
 from itertools import chain
 from math import ceil
 from random import SystemRandom
-from typing import Any, AnyStr, Generator, Iterable, Iterator, List, \
-  MutableSequence, Optional, Text, Type, TypeVar, Union
```
todofixthis: Could we put this back in, so that type hints work correctly?

iota/types.py
```diff
@@ -36,830 +35,829 @@
...
-  def from_string(cls, *args, **kwargs):
-    """
-    Deprecated; use :py:meth:`from_unicode` instead.
+    @classmethod
+    def from_bytes(cls,  # Type[T]
+                   bytes_,  # Union[binary_type, bytearray]
+                   codec=AsciiTrytesCodec.name,  # Text
+                   *args,  # *Any
+                   **kwargs  # **Any
+                   ):  # -> T
```
todofixthis: Type hints should start with `# type: `, and the final one should take the form `# type: (...) -> T` (see [PEP-484](https://www.python.org/dev/peps/pep-0484/#suggested-syntax-for-python-2-7-and-straddling-code) for more info).

iota/types.py
```diff
@@ -36,830 +35,829 @@
...
+                        cls=type(self).__name__
+                    )),
+                context={
+                    'trytes': trytes,
+                },
+            )
+
+        if isinstance(trytes, TryteString):
+            if type(trytes) is TryteString or issubclass(
+              type(trytes), type(self)):
```
todofixthis: Could we restore `incoming_type`?  It might be unnecessary technically, but we don't need to refactor it for PEP-8 compliance.

This is already a huge PR that touches virtually every file in the codebase; I'd prefer it if we limited the scope of functional changes as much as possible.

iota/types.py
```diff
@@ -36,830 +35,829 @@
...
+    and validating checksums.
+    """
+    LEN = Hash.LEN
+
+    def __init__(self,  # TrytesCompatible
+                 trytes,  # Optional[int]
+                 balance=None,  # Optional[int]
+                 key_index=None,  # Optional[int]
+                 security_level=None
+                 ):  # -> None
```
todofixthis: Type hints need fixing.

iota/types.py
```diff
@@ -36,830 +35,829 @@
...
+        super(Address, self).__init__(trytes, pad=self.LEN)
+
+        self.checksum = None
+        # type: Optional[AddressChecksum]
```
todofixthis: Type hints have to be on the same line as the symbol they are documenting.

HerrMuellerluedenscheid:
Hey @todofixthis thanks for the thorough review! We will finish this until the end of the week. We are a little busy due to a hackathon this week. Taking chances to promote IOTA and this lib a little :smiley:

miili:
Hi Phonix, thanks for the thorough review. However I would rather promote IDE agnostic code. Less is more!
Particularly libraries which realize standards and are hopefully mentioned for broader adoption.

Moreover Pycharm can extract type hints from docstrings:

https://www.jetbrains.com/help/pycharm/using-docstrings-to-specify-types.html

todofixthis:
Thanks guys!  Really appreciate this.

WRT type hints, I'm a bit confused because I thought we were following a Python standard — PEP-484.  Is the issue here that other IDEs don't understand those type hints?

The docstring-style type hints are supported in PyCharm, but it's labelled as a [legacy feature](https://www.jetbrains.com/help/pycharm/type-hinting-in-pycharm.html#legacy).

If it turns out that most IDEs do not support PEP-484 but do support docstring type hints, then I agree we should change the code to be more compatible (and no, don't worry; I won't insist on having BOTH docstring and PEP-484 type hints 😹) — but if we go down this route, I'd prefer it if we do that in a separate PR, just to limit the scope of changes.

scottbelden:
I would like add a vote for PEP484 type hints (not docstring hints). I would also recommend using something like `mypy` to have the type hint checks as part of the test suite rather than relying on IDEs because 1) not everyone uses one 2) even if they use one, it might not have hint support and 3) the IDE doesn't enforce it like a failure in the CI would.

Typically I like to have a "test" script that invokes `flake8`, `mypy`, the actual tests, etc and any failures in any of those would cause a CI failure. Basically something like this:

```
#!/bin/bash
set -e

echo "running flake8"
flake8 PROJECT

echo "running mypy"
mypy --ignore-missing-imports --strict PROJECT

echo "running pytest"
python -m coverage run --source PROJECT -m pytest -v $@

python -m coverage report -m
```

This has treated me well and catches problems faster than relying on users to have the right IDE set up.

todofixthis:
I ❤️ the idea of adding style/standards checks to the build+test process.  [I added one a little while ago](https://github.com/iotaledger/iota.lib.py/commit/935dc9a307229f1de69587ab55b571b107bcfa96#diff-354f30a63fb0907d4ad57269548329e3R9) to check for problems in the setuptools project metadata, and I'd be totes keen to expand on that.

todofixthis:
Hey guys.  Really appreciate your hard work on this!  This branch has drifted quite a bit, so I'd like to close this PR and propose a new course of action.

I've also started working on progressively migrating everything over to PEP-8; I should have that wrapped up by the end of the week.  Once that is wrapped up, let's switch focus to adding flake8 (and maybe mypy?) checks to `.travis.yml` and cleaning up any outstanding formatting issues, and then we can finally close #145 😺 

I hate to discard all the hard work you've put into reformatting the code for PEP-8, but I think that the sheer scope of that undertaking was making the process take a lot longer than anybody wanted it to.  So, I'll tackle the boring stuff (reformatting the code), and if you're up for it, you can take on the interesting stuff (setting up automatic checks during the build process and addressing tricky warnings).

Issue #99, opened 2017-11-17
Implement `find_transaction_objects` command


todofixthis:
The JS lib has a [command called `findTransactionObjects`](https://github.com/iotaledger/iota.lib.js/blob/v0.4.4/lib/api/api.js#L410-L431).

This is basically just a wrapper for `findTransactions` that runs the resulting hashes through `getTrytes` and `transactionObject`.

PyOTA has basically the same thing at `iota.commands.extended.utils.find_transaction_objects`; all we need to do is convert this into a proper Command class, and add some unit tests.

lzpap:
Thanks for the implementation @pdecol !

PR #150, opened 2018-01-26
Accept hashes argument to find_transaction_objects

eukaryote31:
Sometimes, the user may already have a list of hashes, in which case we can bypass find_transactions.


todofixthis:
Hey @eukaryote31 thanks for proposing this change!  Unfortunately we can't accept this PR, but #99 may be right up your alley (:

PR #155, opened 2018-02-14
Fixed Array Bounds for Fragment

phil-zxx:
Changed array bound from FRAGMENT_LENGTH (which is 2187 in trytes) to HASH_LENGTH (which is 243 in trits).

### DIFF
--- a/iota/crypto/types.py
+++ b/iota/crypto/types.py
@@ -189,8 +189,8 @@ class PrivateKey(TryteString):
       sponge.absorb(key_fragment)
       sponge.squeeze(hash_trits)
 
-      fragment_start  = i * FRAGMENT_LENGTH
-      fragment_end    = fragment_start + FRAGMENT_LENGTH
+      fragment_start  = i * HASH_LENGTH
+      fragment_end    = fragment_start + HASH_LENGTH
 
       digest[fragment_start:fragment_end] = hash_trits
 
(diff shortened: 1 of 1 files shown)


todofixthis:
Hey @phil-zxx thank you for your PR!

That's a really interesting change; I'm curious to see the unit test that repros the issue.  The Python implementation has diverged a titch from the other API libraries (mostly because of language-specific optimisations), but depending on the issue, it may have implications for the other projects.

phil-zxx:
The variable **hash_trits** is a hash and should always have length 243 (in trits), also indicated by the indices **hash_start** & **hash_end** which imply an array length of **HASH_LENGTH=243** for **hash_trits** (see lines 170-172). So when stitching together the digest, each fragment hash should be treated as having length **HASH_LENGTH**, and not **FRAGMENT_LENGTH**.

Analogously, say you have an array **a = [7,5,2,1]** of length 4 and do the following:
1) **a[100:200] = [9,5]** with both start & end indices completely out of bound, or
2) **a[4:6] = [9,5]** with indices correctly aligned

Python doesn't throw an error in either case, and just appends **[9,5]** to the current array, giving you **a = [7,5,2,1,9,5]**.

My fix basically changes the current implementation from case 1 to case 2. That is, no change in behaviour at all, but correct indexing.

todofixthis:
Ah.  I think I see what you're getting at now.  I'll have to do some more experimentation, but I think you're correct.

One request:  Could we also rename the `fragment_start` and `fragment_end` variables?  Just so that another developer doesn't come along later and try to "fix" it back to how it was (:

phil-zxx:
Added your suggested change.

Issue #157, opened 2018-02-15
to add wereAddressesSpentFrom to PyOTA


lunfardo314:
new API call in JS https://iota.readme.io/v1.3.0/reference#wereaddressesspentfrom

todofixthis:
Scheduled for release: 2.0.5
Thanks @scottbelden !!

PR #158, opened 2018-02-15
add were_addresses_spent_from api

scottbelden:
Resolves https://github.com/iotaledger/iota.lib.py/issues/157

A manual test looks like it's doing the right thing:

```python
from iota import Iota

address = 'P9HZ9OCZONJFIKXVBJLTTGGJPIVX9ACRZEUNYFVSAWYPIMPJVKN9XNKXGUJSYE9NAMLPOGWOCEXPBGIKD'

api = Iota(uri, seed)
r = api.were_addresses_spent_from([address])
print(r)
```

```
{'states': [False], 'duration': 0}
```

### DIFF
--- a/iota/api.py
+++ b/iota/api.py
@@ -417,6 +417,22 @@ class StrictIota(with_metaclass(ApiMeta)):
     """
     return core.StoreTransactionsCommand(self.adapter)(trytes=trytes)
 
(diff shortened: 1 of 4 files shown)


scottbelden:
Good catch! Fixed.

PR #167, opened 2018-03-15
TryteString.random() use cls.LEN as its default length, when available

### LINKED ISSUE #166: TryteString.random() should use cls.LEN as its default length, when available
When calling a class' `random()` method, the `length` parameter should only be required if the class has no `LEN` attribute.

Example:

```python
>>> from iota import Address
>>> Address.random()

# Expected Result
Address(b'... 81 random trytes ...')

# Actual Result
TypeError: random() missing 1 required positional argument: 'length'
```

rpitonak:
Fix #166

### DIFF
--- a/iota/types.py
+++ b/iota/types.py
@@ -49,7 +49,7 @@ class TryteString(JsonSerializable):
   IMPORTANT: A TryteString does not represent a numeric value!
   """
   @classmethod
-  def random(cls, length):
+  def random(cls, length=None):
     # type: (int) -> TryteString
     """
     Generates a random sequence of trytes.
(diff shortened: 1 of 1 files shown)


iota/types.py
```diff
@@ -60,6 +60,12 @@ def random(cls, length):
     alphabet  = list(itervalues(AsciiTrytesCodec.alphabet))
     generator = SystemRandom()
 
+    try:
+      length = length or cls.LEN
```
todofixthis: This would also trip if `length == 0`; could we change this to an explicit `is None` check?

iota/types.py
```diff
@@ -60,6 +60,12 @@ def random(cls, length):
...
+    except AttributeError:  # class has no LEN attribute
+      if length is None:
+        raise TypeError("random() missing 1 required positional argument: 'length'")
```
todofixthis: I feel like this might be confusing to some users, since the method signature shows `length` as optional.

Maybe we could make this error explain that `{class}` does not define a length property.

todofixthis:
Hi @rpitonak let me know if there's anything I can do to help with this!

todofixthis:
Thanks @rpitonak !  I will take a look at this over the weekend.

todofixthis:
Cheers @rpitonak !  The changes look good.  Two last things:

- There's a few conflicts that need to be resolved before we can merge.  We're in the process of PEP-8-ifying the codebase, so it should just be some whitespace changes.
- Could we also include a couple of unit tests for this change?

lzpap:
Change implemented in another PR, see #286 for details.

PR #169, opened 2018-03-28
Corrected parameter-name 'start' to 'index'.

rckey:


todofixthis:
Hey RK.  Thanks for your pull request!

This change potentially breaks backwards compatibility for existing applications, so if we include this, I think it should wait until we're ready to release v2.1.

I do like the idea of keeping the parameter names consistent though.  Could you add an issue to the [PyOTA Issue Tracker](https://github.com/iotaledger/iota.lib.py/issues/) so that we can keep track of it?

rckey:
Done.

PR #173, opened 2018-03-31
security_level optional argument added to get_inputs

### LINKED ISSUE #172: Add security_level parameter to get_inputs, prepare_transfer and send_transfer
is there a reason why `get_inputs` always uses default security level to generate addresses? Maybe it makes sense to add parameter `security_level` for the sake of consistency? otherwise  `get_inputs` is useless for  `security_level`  1 and 3 .

lunfardo314:
fixes #172 

I'm new to github and not familiar with many things in PyOTA code.  Not sure if I'm doing it correctly.
If that is ok, I'll try to do the same for `prepare_transfer` and `send_transfer`


test/commands/extended/get_inputs_test.py
```diff
@@ -90,6 +93,7 @@ def test_pass_optional_parameters_excluded(self):
         'start':      0,
         'stop':       None,
         'threshold':  None,
+        "securityLevel": 2,
```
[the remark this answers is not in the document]
lunfardo314: ok

iota/commands/extended/get_inputs.py
```diff
@@ -38,13 +38,14 @@ def _execute(self, request):
     seed      = request['seed'] # type: Seed
     start     = request['start'] # type: int
     threshold = request['threshold'] # type: Optional[int]
+    security_level = request['securityLevel'] # type: Optional[int]
 
     # Determine the addresses we will be scanning.
     if stop is None:
       addresses =\
         [addy for addy, _ in iter_used_addresses(self.adapter, seed, start)]
```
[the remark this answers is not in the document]
lunfardo314: sure, somehow missed it

lunfardo314:
now I have to figure out how to modify that pull request. Should I create a new one after new commit in my fork?

iota/commands/extended/get_inputs.py
```diff
@@ -38,13 +38,14 @@ def _execute(self, request):
...
     else:
-      addresses = AddressGenerator(seed).get_addresses(start, stop)
+      addresses = AddressGenerator(seed, security_level).get_addresses(start, stop)
```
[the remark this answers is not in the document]
lunfardo314: stuck with this one.  `test.commands.extended.get_inputs_test:GetInputsCommandTestCase.test_start` patches method with mock function, while in case of test for default use of `security_level`  it doesn't work like that. Whole `AddressGenerator` class must be patched (or at least I don't know another obvious way). 
Don't know how correctly to patch the class because when I do that (with subclass) all calls to `super` goes crazy. Ideas?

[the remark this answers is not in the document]
todofixthis: Ah.  Sorry; I should have paid more attention to the test body.

Probably the most straightforward way to do it is to write a test that calls `GetInputsCommand` with `security_level=1` (or `3`) and verify that the resulting addresses are correct.

Admittedly, the scope would be a too broad – since it also covers `AddressGenerator` functionality – but it's fairly straightforward to write, and it should make the test easier to understand.

todofixthis:
Keep on committing to the same branch, and GitHub will automatically add them to the pull request (:

iota/commands/extended/get_inputs.py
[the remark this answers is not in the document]
lunfardo314: I learned a lot trying to do that! Yes, this is straightforward but tests would be slower. Another (non elegant) way would be to put some empty hook into __init__ of AddressGenerator and then patch it. Or introduce some global function like get_address_generator_class of get_address_generator (lots of refactoring). That's for tomorrow, now I go to sleep :)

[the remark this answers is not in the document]
todofixthis: True, but that could cause more headaches down the road; give the [TDD series on Enterprise Craftsmanship](http://enterprisecraftsmanship.com/2015/06/29/test-induced-design-damage-or-why-tdd-is-so-painful/) a read for more info.

I'm OK with slow unit tests — the crypto tests already take up the bulk of it (especially if the optional C extension is not installed), so another few milliseconds to generate an address or two is just a drop in the ocean.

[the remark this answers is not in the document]
lunfardo314: so I created a test for all 3 level's of security, explicitly checking 3 addresses with static seed. It doesn't take very long. 
Thanks for the link above, it does make sense!

iota/commands/extended/utils.py
```diff
@@ -37,17 +37,20 @@ def find_transaction_objects(adapter, **kwargs):
     return []
 
 
-def iter_used_addresses(adapter, seed, start):
+def iter_used_addresses(adapter, seed, start, security_level=None):
   # type: (BaseAdapter, Seed, int) -> Generator[Tuple[Address, List[TransactionHash]]]
```
todofixthis: Could we update the type hint for the method as well?

todofixthis:
Thanks @lunfardo314 !  I will review your latest changes later in the week.

I think we're definitely on the right track here, though.  If you want, you could build on this to integrate `security_level` into `prepare_transfer` and `send_transfer`, and I think that would neatly wrap up #172.

lunfardo314:
@todofixthis please look here https://github.com/iotaledger/iota.lib.py/blob/08230b0d836cbf365c561631a69e55e869473f3f/iota/commands/extended/get_inputs.py#L47 
I guess it must be `addresses = AddressGenerator(seed).get_addresses(start, stop-start)` .
I can't cover it with a test right now because I'm still trying to understand the code and tests, how it works and how to write them :) Planning to do `prepare_transfer` and `send_transfer` too but writing test will take time

todofixthis:
Thanks for continuing to work on this!  I'm at a conference all week, so I won't be able to give this proper attention until next week.  Just wanted to let you know I haven't forgotten about it, and I'm looking forward to checking it out 😺

iota/commands/extended/prepare_transfer.py
```diff
@@ -130,6 +133,7 @@ def __init__(self):
 
         # Optional parameters.
         'changeAddress': Trytes(result_type=Address),
+        'securityLevel': f.Choice([1, 2, 3]) | f.Optional(default=AddressGenerator.DEFAULT_SECURITY_LEVEL),
```
todofixthis: ❔ Could we make this match the filters for the other commands?  This will ensure that the error codes/messages for invalid inputs remain consistent.

test/commands/extended/get_inputs_test.py
```diff
@@ -818,3 +867,107 @@ def mock_address_generator(ag, start, step=1):
...
+    input0 = response['inputs'][0]
+    input1 = response['inputs'][1]
+    self.assertIsInstance(input0, Address)
+    self.assertIsInstance(input1, Address)
+    self.assertEqual(input0, self.addy1)
+    self.assertEqual(input1, self.addy2)
+    self.assertEqual(input0.balance, 11)
+    self.assertEqual(input0.key_index, 1)
+    self.assertEqual(input1.balance, 11)
+    self.assertEqual(input1.key_index, 2)
```
todofixthis: ❔ Could we make these lines easier to read by reordering them, so that each address is checked separately?

test/commands/extended/get_inputs_test.py
```diff
@@ -818,3 +867,107 @@ def mock_address_generator(ag, start, step=1):
...
+        yield addy
+
+    with mock.patch(
+        'iota.crypto.addresses.AddressGenerator.create_iterator',
+        mock_address_generator,
+    ):
+      with mock.patch(
+        'iota.commands.core.GetBalancesCommand._execute',
+        mock_get_balances_execute,
+      ):
```
todofixthis: 💡 If you'd prefer to avoid mocking private methods, here's the way the other unit tests for `GetInputsCommand` simulate API responses:

```python
def test_start(self):
  ...
    self.adapter.seed_response('getBalances', {
      'balances': [86],
    })
    ...
```

`GetInputsCommandTestCase.setup()` injects a `MockAdapter` into the command object, which allows you to specify what response it should simulate for different commands.

test/commands/extended/get_inputs_test.py
```diff
@@ -818,3 +867,107 @@ def mock_address_generator(ag, start, step=1):
...
+
+    # one address with index 0 for selected security levels for the random seed.
+    # to check with respective outputs from command
+    seed = Seed.random()
+    security_levels_to_test = [1, 2] # at least one is non-default
+    addrs = {l: AddressGenerator(seed, l).get_addresses(0)[0] for l in security_levels_to_test}
+
+    for securityLevel in security_levels_to_test:
+      for stop in [True, False]:
+        response = invoke_cmd(seed, stop, securityLevel)
```
todofixthis: 🤔 I like what you're going for here, but I am concerned that making the unit test more DRY incurs too high of a cost:

- If any assertion fails, the test stops, which limits the amount of information available for diagnosing test failures.

  For example, suppose the test fails when `stop = False` and `securityLevel = 1`.  As is, we can't be sure if _both_ conditions are required to cause the test failure, or if _either_ will.  If we split this into multiple tests, then we can see all of the different scenarios that cause pass/fail results, which will help us to look for patterns.
- It becomes more difficult to use the test as documentation; developers have to invest significant effort into understanding the unit test, before they can use it to understand the intended behaviour.
- Seeing conditionals and nested loops in a unit test makes me a bit nervous; this is where the complexity of the functionality reaches a point where I'd recommend we add some unit tests — except then we'd have to write tests for our unit tests!

iota/commands/extended/get_inputs.py
```diff
@@ -38,13 +38,14 @@ def _execute(self, request):
...
+    security_level = request['securityLevel'] # type: Optional[int]
 
     # Determine the addresses we will be scanning.
     if stop is None:
       addresses =\
-        [addy for addy, _ in iter_used_addresses(self.adapter, seed, start)]
+        [addy for addy, _ in iter_used_addresses(self.adapter, seed, start, security_level=security_level)]
     else:
-      addresses = AddressGenerator(seed).get_addresses(start, stop)
+      addresses = AddressGenerator(seed, security_level).get_addresses(start, stop - start)
```
todofixthis: 👍 Good catch; thanks for fixing the `count` parameter, too!

iota/api.py
```diff
@@ -940,6 +954,11 @@ def send_trytes(self, trytes, depth, min_weight_magnitude=None):
 
       If not provided, a default value will be used.
 
+    :param security_level:
+      Number of iterations to use when generating new addresses (see get_new_addresses).
+      This value must be between 1 and 3, inclusive.
+      If not set, defaults to AddressGenerator.DEFAULT_SECURITY_LEVEL = 2
```
todofixthis: ❔ I think this was meant to be added to the docstring for `send_transfer`, not `send_trytes`.

iota/commands/extended/prepare_transfer.py
lunfardo314: ok, I'll make it with Min and Max. Thought that `security_level` semantically is more an enumeration than range of integers :)

test/commands/extended/get_inputs_test.py
lunfardo314: ok

test/commands/extended/get_inputs_test.py
lunfardo314: I'd prefer to avoid mocking private methods and I understand the method of doing it with `seed_response`. In this case the goal was to make getBalances respond dynamically depending on request parameter: `stop and start` -> number of addresses generated -> number of balances returned, because originally it was testing several combinations of start and stop.
Changed to more static version by seeding response [11,11]

iota/api.py
lunfardo314: oops...

test/commands/extended/get_inputs_test.py
lunfardo314: actually, I agree with you: is is unnecessary complication which makes me nervous too :) Replaced by two unit tests, both for non-default `security_level` value

lunfardo314:
yay! Glad to contribute even a little bit

PR #177, opened 2018-05-11
Fixes invalid receive address in examples

msti:
The receive address contains an invalid character: **2**

This pull request replaces the invalid character **2** with **B**


todofixthis:
Good catch, thanks Mike!!

PR #189, opened 2018-06-27
usage of bundle hash by find_transactions

### LINKED ISSUE #186: find_transactions requires tx hashes on iota.lib.py but it should require bundle hashes
https://iota.readme.io/reference#findtransactions

According the reference this call should require list of bundle hashes not Transaction hashes

``` 
   :param bundles:
      List of transaction IDs.
```

oskyk:
fixes #186


todofixthis:
Hi @oskyk thanks for submitting this fix!  Looks great; I'll merge it now.  Thanks!!

Issue #195, opened 2018-07-06
Add default value for `depth` parameter


todofixthis:
Word on the street is, the optimal value for `depth` is 3 <sup>\[citation needed\]</sup>.

Developers who aren't familiar with how the Tangle works don't really know what value to use for `depth`, so we should probably provide a sensible default.

Make `depth` default to 3 for the following commands:

- `promote_transaction`
- `replay_bundle`
- `send_transfer`
- `send_trytes`

todofixthis:
Scheduled for release: 2.0.7
Thanks @redondo-mk!

PR #200, opened 2018-07-18
Setting default value for depth to 3

redondo-mk:
Fix #195 . Please review and let me know if I missed something.


todofixthis:
Awesome, thanks @redondo-mk !!  I will take a look this weekend 😺

PR #205, opened 2018-09-05
Create decode_bundle_messages.py

harleylang:
I really struggled in writing code that grabs messages across transactions in a bundle, so here is a mwe that others may find beneficial.

### DIFF
--- a/examples/decode_bundle_messages.py
+++ b/examples/decode_bundle_messages.py
@@ -1,0 +1,23 @@
+from iota import Iota
+from iota import TryteString
+from iota.commands.extended.utils import get_bundles_from_transaction_hashes as goget
+
+# declare node; change as you see fit
+api = Iota('https://nodes.thetangle.org:443')
+
+# this MWE assumes that you already have you already have a bundle hash; one is declared here as an example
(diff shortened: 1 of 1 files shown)


todofixthis:
Hi @harleylang thanks for submitting this!

Have a look at `Bundle.get_messages()` — it might be possible to make the code even simpler by using this method.

harleylang:
Hey, thanks for your feedback! I was unaware of that function at the time of my OP and I appreciate the reference.

I have been giving this a lot of thought and have tested a few options for message gathering over the past few months. So far I have found that the most efficient message-gathering function pings a node for tangle information only as needed. `Bundle.get_messages()' pings the node for info on each recursion. Larger messages take exponentially more time to recover with this method. In my next comment to this thread, I am going to propose a function that cuts-down node/tangle pings by gathering transaction trytes at the onset. Next, the function iterates through that information for the message. From my testing, that seems to be the most efficient way to gather messages. For a 32-index message, this cuts the duration down from ~25 seconds to 5 seconds. 

After I post my solution, I welcome peer review to ensure that we identify and accept the function that is the most efficient, but also safe. (The latter of which is not currently within my expertise.)

harleylang:
Further to my last comment, I want to clarify that the ping issue is not /specifically/ with `Bundle.get_messages()` but instead with `Iota.get_bundles()` which I understand needs to be used to generate a bundle object and therefore call `bundleobj.get_messages()`.

I will post comparison code and function options soon.

PR #236, opened 2019-09-21
Update the core commands to match the latest IRI 1.8.1 release

### LINKED ISSUE #235: Update the core commands to match the latest IRI 1.8.1 release
The core commands implement the API endpoints which are provided by IRI and are contained in the `StrictIota` class. I compared the current implementation with the latest IRI 1.8.1 release (https://github.com/iotaledger/iri/blob/v1.8.1-RELEASE/src/main/java/com/iota/iri/service/API.java). The following changes are needed to get `StrictIota` up to date. 

Missing commands:
- **getNodeAPIConfiguration:** This will be needed for #138
- **getMissingTransactions**

Missing parameters:
- **getBalances**: `tips` is missing as an optional parameter. `references` is new in the response and could be added to the response filter.
- **getTransactionsToApprove:**  `references` is missing as an optional parameter.

Changed fields in the response:
- **getBalances**: `milestone` (type: `Address`) get removed and `milestoneIndex` (type: `int`)  got added to the response. Shall we remove `milestone`? Or keep both to ensure backwards compatibility?

Fields that I would add to the response filter because without these values in the response the API call is useless:
- **wereAddressesSpentFrom**:  `states`
- **checkConsistency:** `state`

Furthermore for all commands :
- In the source code are references to https://iota.readme.io/docs/  As this website is not used anymore this should be updated to respective references in the official docs https://docs.iota.org/docs/node-software/0.1/iri/references/api-reference

Let me know what you think. I planning to implement this myself and I will open a pull request soon.

pdecol:
resolves #235

### DIFF
--- a/iota/commands/core/check_consistency.py
+++ b/iota/commands/core/check_consistency.py
@@ -5,7 +5,7 @@ from __future__ import absolute_import, division, print_function, \
 import filters as f
 
 from iota import TransactionHash
-from iota.commands import FilterCommand, RequestFilter, ResponseFilter
+from iota.commands import FilterCommand, RequestFilter
 from iota.filters import Trytes
 
 __all__ = [
@@ -25,7 +25,7 @@ class CheckConsistencyCommand(FilterCommand):
         return CheckConsistencyRequestFilter()
 
     def get_response_filter(self):
-        return CheckConsistencyResponseFilter()
+        pass
 
 
 class CheckConsistencyRequestFilter(RequestFilter):
@@ -35,11 +35,4 @@ class CheckConsistencyRequestFilter(RequestFilter):
                 f.Required |
                 f.Array |
                 f.FilterRepeater(f.Required | Trytes(TransactionHash)),
-        })
-
-
-class CheckConsistencyResponseFilter(ResponseFilter):
-    def __init__(self):
-        super(CheckConsistencyResponseFilter, self).__init__({
-            'state': f.Required | f.Type(bool),
         })
--- a/iota/commands/core/check_consistency.py
+++ b/iota/commands/core/check_consistency.py
@@ -5,7 +5,7 @@ from __future__ import absolute_import, division, print_function, \
 import filters as f
 
 from iota import TransactionHash
-from iota.commands import FilterCommand, RequestFilter
+from iota.commands import FilterCommand, RequestFilter, ResponseFilter
 from iota.filters import Trytes
 
 __all__ = [
@@ -25,7 +25,7 @@ class CheckConsistencyCommand(FilterCommand):
         return CheckConsistencyRequestFilter()
 
     def get_response_filter(self):
-        pass
+        return CheckConsistencyResponseFilter()
 
 
 class CheckConsistencyRequestFilter(RequestFilter):
@@ -35,4 +35,11 @@ class CheckConsistencyRequestFilter(RequestFilter):
                 f.Required |
                 f.Array |
                 f.FilterRepeater(f.Required | Trytes(TransactionHash)),
+        })
+
+
+class CheckConsistencyResponseFilter(ResponseFilter):
+    def __init__(self):
+        super(CheckConsistencyResponseFilter, self).__init__({
+            'state': f.Required | f.Type(bool),
         })
--- a/iota/commands/core/get_balances.py
+++ b/iota/commands/core/get_balances.py
@@ -3,8 +3,9 @@ from __future__ import absolute_import, division, print_function, \
     unicode_literals
 
 import filters as f
+from six import iteritems
 
-from iota import Address
+from iota import TransactionHash
 from iota.commands import FilterCommand, RequestFilter, ResponseFilter
 from iota.filters import AddressNoChecksum, Trytes
 
@@ -44,12 +45,37 @@ class GetBalancesRequestFilter(RequestFilter):
                     f.Min(0) |
                     f.Max(100) |
                     f.Optional(default=100),
+
+                'tips':
+                    f.Array | f.FilterRepeater(
+                        f.Required |
+                        Trytes(TransactionHash) |
+                        f.Unicode(encoding='ascii', normalize=False),
+                    )
             },
 
             allow_missing_keys={
-                'threshold',
+                'threshold', 'tips',
             },
         )
+
+    def _apply(self, value):
+        value = super(GetBalancesRequestFilter, self)._apply(
+            value
+        )  # type: dict
+
+        if self._has_errors:
+            return value
+
+        # Remove null search terms.
+        # Note: We will assume that empty lists are intentional.
+        search_terms = {
+            term: query
+            for term, query in iteritems(value)
+            if query is not None
+        }
+
+        return search_terms
 
 
 class GetBalancesResponseFilter(ResponseFilter):
@@ -57,6 +83,12 @@ class GetBalancesResponseFilter(ResponseFilter):
         super(GetBalancesResponseFilter, self).__init__({
             'balances': f.Array | f.FilterRepeater(f.Int),
 
-            'milestone':
-                f.ByteString(encoding='ascii') | Trytes(Address),
+            'milestoneIndex': f.Int,
+
+            'references':
+                f.Array | f.FilterRepeater(
+                    f.Required |
+                    Trytes(TransactionHash) |
+                    f.Unicode(encoding='ascii', normalize=False),
+                ),
         })
--- a/iota/commands/core/get_balances.py
+++ b/iota/commands/core/get_balances.py
@@ -83,8 +83,6 @@ class GetBalancesResponseFilter(ResponseFilter):
         super(GetBalancesResponseFilter, self).__init__({
             'balances': f.Array | f.FilterRepeater(f.Int),
 
-            'milestoneIndex': f.Int,
-
             'references':
                 f.Array | f.FilterRepeater(
                     f.Required |
--- a/iota/commands/core/get_balances.py
+++ b/iota/commands/core/get_balances.py
@@ -85,8 +85,7 @@ class GetBalancesResponseFilter(ResponseFilter):
 
             'references':
                 f.Array | f.FilterRepeater(
-                    f.Required |
-                    Trytes(TransactionHash) |
-                    f.Unicode(encoding='ascii', normalize=False),
+                    f.ByteString(encoding='ascii') |
+                    Trytes(TransactionHash)
                 ),
         })
--- a/iota/commands/core/get_missing_transactions.py
+++ b/iota/commands/core/get_missing_transactions.py
@@ -1,0 +1,47 @@
+# coding=utf-8
+from __future__ import absolute_import, division, print_function, \
+    unicode_literals
+
+import filters as f
+
+from iota import TransactionHash
+from iota.commands import FilterCommand, RequestFilter, ResponseFilter
+from iota.filters import Trytes
+
+__all__ = [
+    'GetMissingTransactionsCommand',
+]
+
+
+class GetMissingTransactionsCommand(FilterCommand):
+    """
+    Executes `getMissingTransactions` command.
+
+    See :py:meth:`iota.api.StrictIota.get_missing_transactions`.
+    """
+    command = 'getMissingTransactions'
+
+    def get_request_filter(self):
+        return GetMissingTransactionsRequestFilter()
+
+    def get_response_filter(self):
+        return GetMissingTransactionsResponseFilter()
+
+
+class GetMissingTransactionsRequestFilter(RequestFilter):
+    def __init__(self):
+        # ``getMissingTransactions`` does not accept any parameters.
+        # Using a filter here just to enforce that the request is empty.
+        super(GetMissingTransactionsRequestFilter, self).__init__({})
+
+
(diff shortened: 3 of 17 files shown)


pdecol:
I usually like it to clean up code that I touch. Since making the code PEP8 conform is still work in progress I changed the indentation level to 4 spaces in existing test files where I added more tests. That is the reason why this pull request has quite a lot of line changes. If its prefered to have these changes in a separate ticket/pull request let me know and I can change it back to 2 spaces.

iota/commands/core/get_balances.py
```diff
@@ -44,19 +45,50 @@ def __init__(self):
...
+        return search_terms
+
 
 class GetBalancesResponseFilter(ResponseFilter):
     def __init__(self):
         super(GetBalancesResponseFilter, self).__init__({
             'balances': f.Array | f.FilterRepeater(f.Int),
 
-            'milestone':
-                f.ByteString(encoding='ascii') | Trytes(Address),
```
pdecol: Note that this is a potential backwards incompatible change. `milestone` was removed in IRI version 1.4.1.6 (https://github.com/iotaledger/iri/blob/v1.4.1.6/src/main/java/com/iota/iri/service/dto/GetBalancesResponse.java) which was released on the 6th of January 2018. Since this was more than one and a half years ago I think it is justified to remove it here too.

todofixthis:
Hi @pdecol thanks for working on this!  Really appreciate the attention to detail and being proactive about keeping the codebase consistent!

Will have a look at this today 👍🏻

iota/commands/core/get_missing_transactions.py
```diff
@@ -0,0 +1,47 @@
...
+    def __init__(self):
+        # ``getMissingTransactions`` does not accept any parameters.
+        # Using a filter here just to enforce that the request is empty.
+        super(GetMissingTransactionsRequestFilter, self).__init__({})
+
+
+class GetMissingTransactionsResponseFilter(ResponseFilter):
+    def __init__(self):
+        super(GetMissingTransactionsResponseFilter, self).__init__({
+            'hashes':
```
lzpap: hey, I have a question about response filters. Why is the duration field never filtered in responses?

test/commands/core/get_node_api_configuration_test.py
```diff
@@ -0,0 +1,95 @@
...
+                'foo': 'bar',
+            },
+
+            {
+                'foo': [f.FilterMapper.CODE_EXTRA_KEY],
+            },
+        )
+
+
+class GetNodeAPIConfigurationFilterTestCase(BaseFilterTestCase):
```
lzpap:
```suggestion
class GetNodeAPIConfigurationResponseFilterTestCase(BaseFilterTestCase):
```

iota/commands/core/get_missing_transactions.py
todofixthis: Good question.  The approach I went for with these was to be restrictive towards requests (because those are within the developer's control) and permissive towards responses (because those are up to the IRI node and out of the developer's control).

Response filters generally only do basic conversions (e.g., for addresses, transaction hashes, etc.); I try to avoid situations where a response from the node would generate an exception in the response filter (principle of least astonishment).

iota/commands/core/get_balances.py
todofixthis: 👍🏻 Sounds good to me.

iota/commands/core/check_consistency.py
```diff
@@ -36,3 +36,10 @@ def __init__(self):
                 f.Array |
                 f.FilterRepeater(f.Required | Trytes(TransactionHash)),
         })
+
+
+class CheckConsistencyResponseFilter(ResponseFilter):
+    def __init__(self):
+        super(CheckConsistencyResponseFilter, self).__init__({
+            'state': f.Required | f.Type(bool),
+        })
```
todofixthis: ❔ Pending https://github.com/iotaledger/iota.lib.py/pull/236/files#r327037008

iota/commands/core/get_missing_transactions.py
todofixthis: That said, I can see the other side of the argument — if the response from the node contains missing/unexpected values, that would be more astonishing than if the lib raised an exception \[that could be caught and handled accordingly in the app\].

For the same reason, it's never really sat well with me that `get_bundles()` doesn't make the validation optional (done this way to match the behaviour of the JS lib at the time, but my preference would be to disable validation by default unless the caller explicitly requests it).

I've never made any of this explicit in the code, so might be good to make a decision one way or the other, and then document it somewhere.

iota/commands/core/get_node_api_configuration.py
```diff
@@ -0,0 +1,45 @@
...
+
+class GetNodeAPIConfigurationResponseFilter(ResponseFilter):
+    def __init__(self):
+        super(GetNodeAPIConfigurationResponseFilter, self).__init__({
+            'maxFindTransactions': f.Type(int),
+            'maxRequestsList': f.Type(int),
+            'maxGetTrytes': f.Type(int),
+            'maxBodyLength': f.Type(int),
+            'testNet': f.Type(bool),
+            'milestoneStartIndex': f.Type(int),
```
todofixthis: ❔ Pending https://github.com/iotaledger/iota.lib.py/pull/236/files#r327037008

iota/commands/core/were_addresses_spent_from.py
```diff
@@ -37,3 +37,14 @@ def __init__(self):
...
+
+class WereAddressesSpentFromResponseFilter(ResponseFilter):
+    def __init__(self):
+        super(WereAddressesSpentFromResponseFilter, self).__init__({
+            'states':
+                f.Array | f.FilterRepeater(
+                    f.Required |
+                    f.Type(bool),
+                ),
+        })
```
todofixthis: ❔ Pending https://github.com/iotaledger/iota.lib.py/pull/236/files#r327037008

test/commands/core/get_missing_transactions_test.py
```diff
@@ -0,0 +1,116 @@
...
+
+import filters as f
+from filters.test import BaseFilterTestCase
+
+from iota import Iota, TransactionHash
+from iota.adapter import MockAdapter
+from iota.commands.core import GetMissingTransactionsCommand
+
+
+class GetMissingTransactionsFilterTestCase(BaseFilterTestCase):
```
todofixthis:
```suggestion
class GetMissingTransactionsRequestFilterTestCase(BaseFilterTestCase):
```

test/commands/core/get_missing_transactions_test.py
```diff
@@ -0,0 +1,116 @@
...
+    filter_type = \
+        GetMissingTransactionsCommand(MockAdapter()).get_request_filter
+    skip_value_check = True
+
+    def test_pass_empty(self):
+        """
+        The incoming response is (correctly) empty.
```
todofixthis:
```suggestion
        The incoming request is (correctly) empty.
```

test/commands/core/get_missing_transactions_test.py
```diff
@@ -0,0 +1,116 @@
...
+        request = {}
+
+        filter_ = self._filter(request)
+
+        self.assertFilterPasses(filter_)
+        self.assertDictEqual(filter_.cleaned_data, request)
+
+    def test_fail_unexpected_parameters(self):
+        """
+        The incoming response contains unexpected parameters.
```
todofixthis:
```suggestion
        The incoming request contains unexpected parameters.
```

test/commands/core/get_node_api_configuration_test.py
```diff
@@ -0,0 +1,95 @@
...
+
+
+class GetNodeAPIConfigurationRequestFilterTestCase(BaseFilterTestCase):
+    filter_type = \
+        GetNodeAPIConfigurationCommand(MockAdapter()).get_request_filter
+    skip_value_check = True
+
+    def test_pass_empty(self):
+        """
+        The incoming response is (correctly) empty.
```
todofixthis:
```suggestion
        The incoming request is (correctly) empty.
```

test/commands/core/get_node_api_configuration_test.py
```diff
@@ -0,0 +1,95 @@
...
+        request = {}
+
+        filter_ = self._filter(request)
+
+        self.assertFilterPasses(filter_)
+        self.assertDictEqual(filter_.cleaned_data, request)
+
+    def test_fail_unexpected_parameters(self):
+        """
+        The incoming response contains unexpected parameters.
```
todofixthis:
```suggestion
        The incoming request contains unexpected parameters.
```

test/commands/core/get_balances_test.py
```diff
@@ -14,290 +14,345 @@
...
+    def test_pass_happy_path_with_tips(self):
+        """
+        Typical invocation of ``getBalances`` with tips.
+        """
+        request = {
+            'addresses': [self.trytes1, self.trytes2],
+
+            'threshold': 80,
+
+            'tips': [self.trytes3, ],
```
todofixthis:
```suggestion
            'tips': [self.trytes3],
```

test/commands/core/get_balances_test.py
```diff
@@ -14,290 +14,345 @@
...
+
+    def test_balances(self):
+        """
+        Typical ``getBalances`` response.
+        """
+        filter_ = self._filter({
+            'balances':       ['114544444', '0', '8175737'],
+            'duration':       42,
+            'milestoneIndex': 128,
+            'references':     [self.trytes1, ]
```
todofixthis:
```suggestion
            'references':     [self.trytes1]
```

test/commands/core/get_balances_test.py
```diff
@@ -14,290 +14,345 @@
...
+
+        self.assertFilterPasses(filter_)
+        self.assertDictEqual(
+            filter_.cleaned_data,
+
+            {
+                'balances':       [114544444, 0, 8175737],
+                'duration':       42,
+                'milestoneIndex': 128,
+                'references':     [self.trytes1, ]
```
todofixthis:
```suggestion
                'references':     [self.trytes1]
```

test/commands/core/were_addresses_spent_from_test.py
```diff
@@ -9,167 +9,208 @@
...
+            filter_.cleaned_data,
+
+            {
+                'states': [True, ]
+            },
+        )
+
+    def test_multiple_addresses(self):
+        """
+        Typical ``wereAddressesSpentFrom`` response for a multiple address
```
todofixthis:
```suggestion
        Typical ``wereAddressesSpentFrom`` response for multiple addresses.
```

lzpap:
> Minor changes requested, but generally looks really good, thanks @pdecol!
> 
> @lzpap we need to come to a decision about whether/how to validate responses coming from the IRI node (survey app developer community and see if there's a clear preference one way or the other, maybe?).

Hey @todofixthis I will start with coordinating with the other libs to see if they validate requests/responses. I am generally for filters in both directions. Then you could handle exceptions and also it might be easier to reveal bugs with future IRI API changes.

lzpap:
> > Minor changes requested, but generally looks really good, thanks @pdecol!
> > @lzpap we need to come to a decision about whether/how to validate responses coming from the IRI node (survey app developer community and see if there's a clear preference one way or the other, maybe?).
> 
> Hey @todofixthis I will start with coordinating with the other libs to see if they validate requests/responses. I am generally for filters in both directions. Then you could handle exceptions and also it might be easier to reveal bugs with future IRI API changes.

@todofixthis I talked with the devs of the other libraries, and the conclusion is that responses are not filtered/validated/checked in general, except bundle and transactions hashes. I think it would be nice to make PyOTA consistent with this behavior.

pdecol:
Thanks for the review! I implemented the requested changes and reverted the response filters for the fields that we decided not to validate.

iota/commands/core/get_balances.py
```diff
@@ -44,19 +45,48 @@ def __init__(self):
...
+            'references':
+                f.Array | f.FilterRepeater(
+                    f.Required |
+                    Trytes(TransactionHash) |
+                    f.Unicode(encoding='ascii', normalize=False),
+                ),
```
todofixthis: ❔ Could we remove `f.Required` from the response filter?  Match behaviour of other libs, so that node responses don't generate exceptions.
❔ The `f.Unicode` filter will convert `TransactionHash` values into strings.  This is desirable for request filters (so that the string values can be converted to JSON to send to the node), but probably counter-productive for response filters (the client would likely prefer to work with `TransactionHash` objects rather than strings).  Could we remove it from the response filter?

Apologies for not picking up on this during the first review.

pdecol:
No worries. Thanks for spotting this mistake. I removed `f.Required` and `f.Unicode` and I added `f.ByteString` so that it is consistent with the response filters in findTransaction, getTrytes, getTips and getMissingTransactions.


One more thing that I would like to ask you. Can you change this URL here in GitHub? (https://iota.readme.io/ -> https://docs.iota.org/) I think this should be the very last URL that needs to be fixed.

image

todofixthis:
LGTM Thanks @pdecol !!

lzpap:
Thanks @pdecol !

PR #242, opened 2019-10-08
Implement getTransactionObjects command

### LINKED ISSUE #241: Missing get transaction objects command
The JS library implements `getTransactionObjects` command, that is essentially the same as `findTransactionObjects`, but awaits transaction hashes as inputs.

To make PyOTA more consistent with other libraries, implement the command.

lzpap:
New command to harmonize PyOTA with JS lib.
Similar to findTransactionObjects, but input
is a list of transaction hashes.

resolves #241

### DIFF
--- a/iota/commands/extended/get_transaction_objects.py
+++ b/iota/commands/extended/get_transaction_objects.py
@@ -1,0 +1,59 @@
+# coding=utf-8
+from __future__ import absolute_import, division, print_function, \
+    unicode_literals
+
+from typing import Iterable, List, Optional
+
+import filters as f
+
+from iota import Transaction, TransactionHash
+from iota.commands.core import GetTrytesCommand
+from iota.commands import FilterCommand, RequestFilter
+from iota.filters import Trytes
+
+__all__ = [
+    'GetTransactionObjectsCommand',
+]
+
+
+class GetTransactionObjectsCommand(FilterCommand):
+    """
+    Executes `GetTransactionObjectsCommand` command.
+
+    See :py:meth:`iota.api.StrictIota.get_transaction_objects`.
+    """
+    command = 'getTransactionObjects'
+
+    def get_request_filter(self):
+        return GetTransactionObjectsRequestFilter()
+
+    def get_response_filter(self):
+        pass
+
+    def _execute(self, request):
+        hashes = request\
+            .get('hashes') # type: Optional[Iterable[TransactionHash]]
+
+        transactions = []
+        if hashes:
+            gt_response = GetTrytesCommand(adapter=self.adapter)(hashes=hashes)
+
+            transactions = list(map(
+                Transaction.from_tryte_string,
+                gt_response.get('trytes') or [],
+            ))  # type: List[Transaction]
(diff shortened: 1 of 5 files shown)


iota/commands/extended/get_transaction_objects.py
```diff
@@ -0,0 +1,59 @@
...
+__all__ = [
+    'GetTransactionObjectsCommand',
+]
+
+
+class GetTransactionObjectsCommand(FilterCommand):
+    """
+    Executes `GetTransactionObjectsCommand` command.
+
+    See :py:meth:`iota.api.StrictIota.get_transaction_objects`.
```
todofixthis:
```suggestion
    See :py:meth:`iota.api.Iota.get_transaction_objects`.
```

(`StrictIota` contains only commands exposed by the IRI node; original inspiration taken from [redis](https://github.com/andymccurdy/redis-py/blob/2.10.6/redis/client.py#L347-L356))

iota/commands/extended/get_transaction_objects.py
```diff
@@ -0,0 +1,59 @@
...
+
+    def get_request_filter(self):
+        return GetTransactionObjectsRequestFilter()
+
+    def get_response_filter(self):
+        pass
+
+    def _execute(self, request):
+        hashes = request\
+            .get('hashes') # type: Optional[Iterable[TransactionHash]]
```
todofixthis:
```suggestion
            .get('hashes') # type: Iterable[TransactionHash]
```

(the `f.Required` filter in the request filter chain ensures that this value cannot be `None`)

test/commands/extended/get_transaction_objects_test.py
```diff
@@ -0,0 +1,111 @@
...
+
+from unittest import TestCase
+
+import mock
+
+from iota import Iota, MockAdapter, Transaction
+from iota.commands.extended import GetTransactionObjectsCommand
+
+
+class GetTransactionObjectsCommandTestCase(TestCase):
```
todofixthis: ❔ Should we also include tests for `GetTransactionObjectsRequestFilter`?

todofixthis: See #243

iota/commands/extended/get_transaction_objects.py
lzpap: Thanks for noticing!

iota/commands/extended/get_transaction_objects.py
```diff
@@ -0,0 +1,56 @@
...
+        return {
+            'transactions': transactions,
+        }
+
+class GetTransactionObjectsRequestFilter(RequestFilter):
+    def __init__(self):
+        super(GetTransactionObjectsRequestFilter, self).__init__({
+            'hashes':
+                StringifiedTrytesArray(TransactionHash) | f.Required
+                ),
```
todofixthis:
```suggestion
```

lzpap: whoops

PR #244, opened 2019-10-10
Snapshot proof commands

### LINKED ISSUE #46: Need documentation explaining how to handle snapshots
After a snapshot, previous transactions are lost.  This means that methods such as `get_account_data`, `get_new_addresses`, `get_transfers`, etc. can no longer identify "used" addresses correctly.

At this time, the only way to properly identify "pre-snapshot" addresses is to keep track of them before the snapshot, then be sure to set the correct starting index (for key generation) when calling API methods.

1. Determine which API commands are impacted by snapshots (basically anything that relies on `iter_used_addresses`).
2. Update the documentation for each of these API commands with more info about how they are impacted by snapshots.

### LINKED ISSUE #175: `get_inputs` checks possible input addresses not the best way
`get_inputs`  behaves very differently depending on `stop` parameter (is it provided or not). That is due to different ways how addresses are generated and iterated. For example right after snapshot that would give two different results.

Suggestion: `iter_used_addresses` to iterate until encounters first address with *balance=0 and not spent as per `were_addresses_spent_from`* as opposed to current criteria of *first address without transactions*.  
In standard cases that will give wallet's total balance and would be similar to result with big enough interval  of `start` and `stop`

### LINKED ISSUE #217: get_account_data() always returns empty result if there are no transactions
**Important:** Co-ordinate this change with the other IOTA API libraries, to ensure behaviour remains consistent.

If there are no transactions for the generated addresses (e.g., after a snapshot), `get_account_data()` always returns an empty result, even if one or more of the generated addresses have a balance.

This behaviour makes sense when specifying `get_account_data(stop=None, ...)` (the address generator would stop at the 0th address because there are no transactions), but it is not intuitive when specifying a value for `stop` (see #212 for an example).

pdecol:
resolves #217, resolves #175, resolves #46 

This PR makes getNewAddresses, getInputs, getTransfers and getAccountData snapshot proof. It also follows the behaviour of the Javascript library more closely but it is not exactly the same. There is a difference in how these commands behave on addresses with balances and without transactions. I opened a ticket in the JavaScript GitHub repository where I documented the behaviour of the JavaScript library on such addresses (https://github.com/iotaledger/iota.js/issues/427). As I stated in this ticket I think that the behaviour of the JavaScript library is in these cases a bit strange and I think that my suggestion that I also implemented in this PR is more logical. There has been no response so far but maybe @lzpap  could help to coordinate how the commands should behave for addresses with balance and without transactions?

### DIFF
--- a/docs/addresses.rst
+++ b/docs/addresses.rst
@@ -17,16 +17,10 @@ any other financial service.
         These performance issues will be fixed in a future version of the library;
         please bear with us!
 
-        In the meantime, if you are using Python 3, you can install a C extension
+        In the meantime, you can install a C extension
         that boosts PyOTA's performance significantly (speedups of 60x are common!).
 
         To install the extension, run ``pip install pyota[ccurl]``.
-
-        **Important:** The extension is not yet compatible with Python 2.
-
-        If you are familiar with Python 2's C API, we'd love to hear from you!
-        Check the `GitHub issue <https://github.com/todofixthis/pyota-ccurl/issues/4>`_
-        for more information.
 
 PyOTA provides two methods for generating addresses:
 
@@ -60,7 +54,9 @@ method, using the following parameters:
    (defaults to 1).
 -  If ``None``, the API will generate addresses until it finds one that
    has not been used (has no transactions associated with it on the
-   Tangle). It will then return the unused address and discard the rest.
+   Tangle, has no balance and was not spent from). This makes the command
+   safe to use even after a snapshot has been taken. It will then return the
+   unused address and discard the rest.
 -  ``security_level: int``: Determines the security level of the
    generated addresses. See `Security Levels <#security-levels>`__
    below.
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -89,8 +89,10 @@ Parameters
    parameter behaves like the ``stop`` attribute in a ``slice`` object;
    the stop index is *not* included in the result.
 
--  If ``None`` (default), then this method will check every address
-   until it finds one without any transfers.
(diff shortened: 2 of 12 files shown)


lzpap:
hey @pdecol, awesome! I'm on it ;)
However, bear in mind that this might be low priority for the other libs that already have account module, which solves the after snapshot issue by making the lib (locally) stateful.

todofixthis (commented):
Gave it a quick glance last night; will give it a proper look-over today.

todofixthis:
Looking at Chris's comment (https://github.com/iotaledger/iota.js/issues/427#issuecomment-540633581), I'm wondering if we should hold off on this change until we implement the account module in PyOTA — it seems like the changes needed to implement that will conflict with what's being proposed in this PR.  @pdecol @lzpap What do you think?

pdecol:
What I try to clarify in the other ticket is only a part of what is proposed in this PR. To be precise it is about whether or not the `getBalances` API call should be added to `getNewAddresses` and `iter_used_addresses`. By doing so the issues that I mentioned in the other ticket would be solved.

The other changes (adding the `wereAddressesSpentFrom` API call to `getNewAddresses` and `iter_used_addresses` and the change in `get_account_data.py`) are currently implemented that way in the JavaScript library and are needed to establish an identical behaviour between the two libraries.

lzpap:
I suggest then that we break up the PR into the two changes that you mentioned @pdecol , and put the first one on hold until we have a concept about account module in pyota.
The second change is indeed needed to establish identical behavior, I agree.
Thanks for the proactive approach @pdecol! :)

Issue #228, opened 2019-08-23
Prepare transaction offline with seed, send prepared tx to node without needing to invoke seed


jo7399:
Currently working on a demo where i have a buyer and seller side. They communicate with http requests and whilst the buyer transfers to the sender the thought is that the buyer shouldn't need to make use of API calls whilst holding the seed for the funds he whishes to transfer. In Pyota I  thought of creating a transaction with prepare_transfer on the buyer side, which is sent to the seller who then invokes send_trytes. But I have encountered the following issues:

1) The buyer needs to invoke an API to prepare_transfer which I want to avoid.
2) The seller needs to buyer's seed to attach the transaction to the tangle (i.e. to send_trytes, because for a value transaction the api needs the seed as an input) which is obvisouly undesireable. I guess one faces the same problem when using different functions/combinations of different functions.

Ideally I would like the buyer to propose/prepare a transfer with his/her seed and then the transfer is sent to the seller who sends the transaction to a node without requiring access to the buyers seed. From what I understand this would be possible in js: https://docs.iota.org/docs/getting-started/0.1/tutorials/send-iota-tokens

Any ideas how I can do this in python?

Thanks for anyone's help in advance. I'm very new to IOTA so I may well have overlooked something very basic

todofixthis:
Hi @jo7399 I apologise that it has taken me so long to address this.

Unfortunately, PyOTA currently requires a connection to a node to complete the PoW process.  #139 proposes to address this, but it has not been implemented yet.

Other than that, however, your thinking is correct — the buyer can `prepare_transfer` offline (a local IRI node is required, but it doesn't actually have to be connected to the Tangle; see [docs](https://pyota.readthedocs.io/en/latest/adapters.html#routingwrapper) for more info), and the seller can then use `send_trytes` to broadcast the signed transactions to the Tangle.

lzpap:
Hey @jo7399 , the PoW process can be done locally starting with [PyOTA 2.1.0](https://github.com/iotaledger/iota.py/releases/tag/2.1.0).

> The buyer needs to invoke an API to prepare_transfer which I want to avoid.

`prepare_transfer` extended api command in PyOTA does call some API to ensure the prepared bundle is okay. Inside the command:

- `get_inputs` is called if no `inputs` is provided to `prepare_transfer`
- `get_balance` is called to check if the provided `inputs` have sufficient balance
[...]

Issue #139, opened 2018-01-15
Add support for performing `attachToTangle` PoW locally (without relying on IRI node)


todofixthis:
The PoW process can be done 100% offline, so the library should provide a mechanism for doing this locally, without having to send an `attachToTangle` request to a node.

Requirements not 100% nailed down yet, but here are some thoughts:

- Probably easiest for developers if we add e.g., a `local_pow` argument to `Iota.__init__()` (maybe `StrictIota.__init__()`?).
    - An alternative approach would be to use an adapter for this (a la <https://pyota.readthedocs.io/en/develop/adapters.html#routingwrapper>), but that would make it harder for developers to discover this feature.
    - Definitely open to other approaches, post a comment if you think there's a more intuitive/usable way to integrate this into the API (:
- Here are some existing implementations:
    - [Java](https://github.com/iotaledger/iota.lib.java/blob/master/jota/src/main/java/cfb/pearldiver/PearlDiverLocalPoW.java)
    - [C](https://github.com/iotaledger/ccurl/tree/master/src/lib) ("pearl_diver")
    - [Javascript](https://github.com/iotaledger/curl.lib.js)

Also, don't forget to update documentation and example scripts (:

plenarius:
For reference here's how the js promoter creates a `localAttachToTangle` - might be helpful. https://github.com/plenarius/iota-promoter/blob/ca8754c3abc476d54b9940de543180fee03710d4/deps/iota.transactionSpammer.js/src/transactionSpammer.js#L152-L262

scottbelden:
I don't really have experience with c extensions in python, but it seems like it would be more robust and faster if the c version could be wrapped and called via a c extension rather than porting to pure python.

todofixthis:
Aye, I agree.  Based on my experiences porting Curl-P to Python, I am not optimistic about how a pure-Python PearlDiver will perform.

[We already have a C extension for Curl-P](https://github.com/todofixthis/pyota-ccurl/), and with a little bit of work, we could probably [turn it into a wrapper for the ccurl library](https://github.com/todofixthis/pyota-ccurl/issues/10) (which itself includes a PearlDiver implementation).

I think that would be a pretty sweet approach (:

chrisoro:
I wouldn't port it to python but rather use the existing one. Could even use it directly by going the cython route..

For the overall feature itself I would like it to be easy to change, like you said. I would recommend the RoutingWrapper though, but just add the possibility to 'route' that call locally to a different function. 

[...]

PR #245, opened 2019-10-14
Support for Local PoW

lzpap:
Related issues: #139 #228 

To perform PoW (attach_to_tangle) locally, BaseAdatper class has a new attribute, local_pow,
that is set to False by default (and therefore it is backward compatible). By declaring a new api instance with local_pow=True, attach_to_tangle calls will be handled by an interface to the ccurl library. The interface can be installed via an extension module to pyota:
```
pip install pyota[pow]
```
Or alternatively through following the steps in the README at:
https://github.com/lzpap/ccurl.interface.py

Or directly from PyPi:
```
pip install pyota-pow
```

~TODO: write tests~

Please share your thoughts about the change :)

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -44,13 +44,12 @@ To install this extension, use the following command::
 
 Optional Local Pow
 ==================
-To perform Proof-of-Work locally without relying on any node,
+To perform proof-of-work locally without relying on a node,
 you can install an extension module called `PyOTA-PoW`_ .
 
 Specifiy the ``local_pow=True`` argument when creating an
 api instance, that will redirect all ``attach_to_tangle``
 API calls to an interface function in the ``pow`` package.
-Returns the same as a node would for the API call.
 
 To install this extension, use the following command::
 
@@ -118,5 +117,5 @@ can also build the documentation locally:
 .. _ReadTheDocs: https://pyota.readthedocs.io/
 .. _official API: https://docs.iota.org/docs/node-software/0.1/iri/references/api-reference
 .. _tox: https://tox.readthedocs.io/
-.. _Ccurl.interface.py: https://github.com/lzpap/ccurl.interface.py
+.. _Ccurl.interface.py: https://github.com/iotaledger/ccurl.interface.py
 .. _PyOTA-PoW: https://pypi.org/project/PyOTA-PoW/
--- a/README.rst
+++ b/README.rst
@@ -42,6 +42,23 @@ To install this extension, use the following command::
 
    pip install pyota[ccurl]
 
+Optional Local Pow
+==================
+To perform Proof-of-Work locally without relying on any node,
+you can install an extension module called `PyOTA-PoW`_ .
+
+Specifiy the ``local_pow=True`` argument when creating an
+api instance, that will redirect all ``attach_to_tangle``
+API calls to an interface function in the ``pow`` package.
+Returns the same as a node would for the API call.
+
+To install this extension, use the following command::
+
+   pip install pyota[pow]
+
+Alternativley you can take a look on the repository
+`Ccurl.interface.py`_  to install Pyota-PoW.
+Follow the steps depicted in the repo's README file.
 
 Installing from Source
 ======================
@@ -101,3 +118,5 @@ can also build the documentation locally:
 .. _ReadTheDocs: https://pyota.readthedocs.io/
 .. _official API: https://docs.iota.org/docs/node-software/0.1/iri/references/api-reference
 .. _tox: https://tox.readthedocs.io/
+.. _Ccurl.interface.py: https://github.com/lzpap/ccurl.interface.py
+.. _PyOTA-PoW: https://pypi.org/project/PyOTA-PoW/
--- a/examples/local_pow.py
+++ b/examples/local_pow.py
@@ -3,7 +3,6 @@ from __future__ import absolute_import, division, print_function, \
 
 import iota
 from pprint import pprint
-from datetime import datetime
 
 # Generate a random seed.
 myseed = iota.crypto.types.Seed.random()
--- a/examples/local_pow.py
+++ b/examples/local_pow.py
@@ -1,0 +1,39 @@
+from __future__ import absolute_import, division, print_function, \
+    unicode_literals
+
+import iota
+from pprint import pprint
+from datetime import datetime
+
+# Generate a random seed.
+myseed = iota.crypto.types.Seed.random()
+# Get an address generator.
+addres_generator = iota.crypto.addresses.AddressGenerator(myseed)
+
+# Instantiate API. Note the `local_pow=True` argument.
+# This will cause PyOTA to do proof-of-work locally,
+# by using the pyota-pow extension package. (if installed)
+# Find it at: https://pypi.org/project/PyOTA-PoW/
+api = iota.Iota("https://nodes.thetangle.org:443",myseed,local_pow=True)
+
+# Generate two addresses
+addys = addres_generator.get_addresses(1, count=2)
+pprint('Generated addresses are:')
+pprint(addys)
+
+# Preparing transactions
+pt = iota.ProposedTransaction(address = iota.Address(addys[0]),
+                              message = iota.TryteString.from_unicode('Tx1: The PoW for this transaction was done by Pyota-Pow.'),
+                              tag     = iota.Tag(b'LOCALATTACHINTERFACE99999'), # Up to 27 trytes
+                              value   = 0)
+
+pt2 = iota.ProposedTransaction(address = iota.Address(addys[1]),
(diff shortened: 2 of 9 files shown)


iota/api.py
```diff
@@ -82,6 +82,7 @@ def __init__(self, adapter, testnet=False):
             adapter = resolve_adapter(adapter)
 
         self.adapter = adapter  # type: BaseAdapter
+        self.adapter.set_local_pow(local_pow)
```
todofixthis: ❔ Putting this in the adapter might be overkill, since it's not really used by the adapter itself; maybe instead do something similar to what we do for `min_weight_magnitude`?

```py
class StrictIota:
    def __init__(self, adapter, testnet=False, local_pow=False):
        self.local_pow = local_pow

    def attach_to_tangle(self, ..., local_pow=None):
        if local_pow is None:
            local_pow = self.local_pow

        return core.AttachToTangleCommand(self.adapter)(
            ....
            local_pow=local_pow,
        )
```

iota/commands/core/attach_to_tangle.py
```diff
@@ -27,6 +27,18 @@ def get_request_filter(self):
...
+    def _execute(self, request):
+        if self.adapter.local_pow is True:
+            from pow import ccurl_interface
+            powed_trytes = ccurl_interface.attach_to_tangle(
+                request['trytes'],
+                request['branchTransaction'],
+                request['trunkTransaction'],
+                request['minWeightMagnitude']
+            )
+            return { 'trytes' : powed_trytes}
```
todofixthis:
```suggestion
            return {'trytes': powed_trytes}
```

README.rst
```diff
@@ -42,6 +42,23 @@ To install this extension, use the following command::
...
+API calls to an interface function in the ``pow`` package.
+Returns the same as a node would for the API call.
+
+To install this extension, use the following command::
+
+   pip install pyota[pow]
+
+Alternativley you can take a look on the repository
+`Ccurl.interface.py`_  to install Pyota-PoW.
+Follow the steps depicted in the repo's README file.
```
todofixthis: ❔ Out of curiosity, could we also use `Ccurl.interface.py` as a replacement for `pyota-ccurl`?

iota/api.py
lzpap: It might be an overkill, but greatly simplifies the handling. The problem is, if `local_pow` is associated with the api class, we have to modify every call to `*Command` that somehow uses `AttachToTangleCommand` internally. These commands are in the extended api, like `send_transfer` and `send_trytes`.
On the other hand, if `local_pow` is associated with the adapter (value still can be passed during api instantiation for simplicity and user friendliness) then all we have to do is modify the `AttachToTangleCommand` only, and it handles all direct or indirect calls correctly. The reason is that within the `execute_` function of a `*Command`, we only have access to the adapter, but not the api class.

README.rst
lzpap: Theoretically I think it is possible, but requires a bit more work on `Ccurl.interface.py`. I'm not sure about how this would work on Windows. Btw, does PyOTA supports Windows at all?

iota/api.py
todofixthis: Makes sense to me.  Good thinking!

README.rst
todofixthis: 🤔 Good question; I don't think I ever explicitly added Windows support, but I don't think there's anything OS-specific in that code, so it probably should work.  At any rate, I never got any complaints from Windows users... well [except for one](https://github.com/todofixthis/pyota-ccurl/issues/22), but that turned out to be `pip`'s fault 😹

examples/local_pow.py
```diff
@@ -0,0 +1,29 @@
+from __future__ import absolute_import, division, print_function, \
+    unicode_literals
+
+import iota
+from pprint import *
+from datetime import *
+myseed = iota.crypto.types.Seed.random()
+addres_generator = iota.crypto.addresses.AddressGenerator(myseed)
+api = iota.Iota("https://nodes.thetangle.org:443",myseed,local_pow=True)
```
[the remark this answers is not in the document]
lzpap: Yep.

test/local_pow_test.py
```diff
@@ -0,0 +1,197 @@
...
+                self.mwm)
+            self.assertTrue(mocked_ccurl.called)
+            self.assertEqual(result['trytes'], self.bundle)
+
+    def test_mock_adapter(self):
+        """
+        Test if local_pow feature works with MockAdapter.
+        """
+        # Note that we need correct return value to pass the
+        # repsonse filter.
```
todofixthis:
```suggestion
        # response filter.
```

test/local_pow_test.py
```diff
@@ -0,0 +1,197 @@
...
+            self.assertEqual(result['trytes'], self.bundle)
+            # And not by mocked pow pkg
+            self.assertNotEqual(result['trytes'], self.ccurl_bundle)
+
+    def test_http_adapter(self):
+        """
+        Test if local_pow feature works with HttpAdapter.
+        """
+        # Note that we need correct return value to pass the
+        # repsonse filter.
```
todofixthis:
```suggestion
        # response filter.
```

test/local_pow_test.py
```diff
@@ -0,0 +1,197 @@
...
+                self.mwm)
+            self.assertTrue(mocked_ccurl.called)
+            self.assertEqual(result['trytes'], self.bundle)
+
+    def test_sandbox_adapter(self):
+        """
+        Test if local_pow feature works with SandboxAdapter.
+        """
+        # Note that we need correct return value to pass the
+        # repsonse filter.
```
todofixthis:
```suggestion
        # response filter.
```

test/local_pow_test.py
```diff
@@ -0,0 +1,197 @@
...
+                self.branch,
+                self.bundle,
+                self.mwm)
+            self.assertTrue(mocked_ccurl.called)
+            self.assertEqual(result['trytes'], self.bundle)
+
+    def test_sandbox_adapter(self):
+        """
+        Test if local_pow feature works with SandboxAdapter.
+        """
```
todofixthis: 💭  Out of curiosity, is the sandbox node still running?  Is it time to remove `SandboxAdapter`?

test/local_pow_test.py
```diff
@@ -0,0 +1,197 @@
...
+                self.mwm)
+            self.assertTrue(mocked_ccurl.called)
+            self.assertEqual(result['trytes'], self.bundle)
+    
+    def test_routing_wrapper(self):
+        """
+        Test if local_pow feature works with RoutingWrapper.
+        """
+        # Note that we need correct return value to pass the
+        # repsonse filter.
```
todofixthis:
```suggestion
        # response filter.
```

test/local_pow_test.py
```diff
@@ -0,0 +1,197 @@
...
+        with patch('pow.ccurl_interface.attach_to_tangle',
+                    MagicMock(return_value=self.bundle)) as mocked_ccurl:
+            # We are trying to redirect `attach_to_tangle` calls to localhost
+            # with a RoutingWrapper. However, if local_pow=true, the pow
+            # request will not reach the adapter, but will be directed to
+            # ccurl interface.
```
todofixthis: 💭 Ah, very interesting; that makes sense.  Should we also modify the documentation for `RoutingWrapper` to call this out?
﻿
﻿﻿https://pyota.readthedocs.io/en/latest/adapters.html#routingwrapper

test/local_pow_test.py
```diff
@@ -0,0 +1,197 @@
...
+                self.branch,
+                self.bundle,
+                self.mwm)
+            # Ccurl interface was not called
+            self.assertFalse(mocked_ccurl.called)
+            # Result is the one returned by MockAdapter
+            self.assertEqual(result['trytes'], self.bundle)
+            # And not by mocked pow pkg
+            self.assertNotEqual(result['trytes'], self.ccurl_bundle)
+
```
todofixthis:
```suggestion
```

(suggest removing extra newlines at end of file)

test/local_pow_test.py
lzpap: Done ;)

test/local_pow_test.py
lzpap: I think I need to be enlightened on this topic. What's the story with "the sandbox node"?

todofixthis: I think it was a couple of years ago now, the IF stood up a special node that would allow clients to outsource the PoW for `attachToTangle`.  It was an asynchronous process; the client would send the request, and it would get a job ID back.  The client would then need to poll the node until the PoW had been finished, and then it could retrieve the trytes ready to broadcast.

Announcement here:
https://blog.iota.org/iota-sandbox-developer-environment-c70c8206fe5f

The source code for the sandbox node itself is here:
https://github.com/iotaledger/sandbox

PR #249, opened 2019-10-22
Implement broadcast_bundle Api Command

### LINKED ISSUE #246: broadcast_bundle command to match JS lib
Increase consistency of pyota with other iota client libs.
The command `broadcast_bundle`: 

> Re-broadcasts all transactions in a bundle given the tail transaction hash. It might be useful when transactions did not properly propagate, particularly in the case of large bundles.

as written in the JS lib api [reference.](https://github.com/iotaledger/iota.js/blob/next/api_reference.md#module_core.broadcastBundle)

lzpap:
closes #246 

As part of the extended api `Iota` class, this function takes a tail transaction hash of a bundle and re-broadcasts the entire bundle.
It does so by fetching and validating the bundle from the Tangle (`GetBundlesCommand`), and calling `BroadcastTransactionsCommand`core api.
Returns the broadcast `bundle` trytes.

### DIFF
--- a/iota/api.py
+++ b/iota/api.py
@@ -596,7 +596,7 @@ class Iota(StrictIota):
             Tail transaction hash of the bundle.
 
         :return:
-            Dict with the following structure:::
+            Dict with the following structure::
 
                 {
                     'trytes': List[TransactionTrytes],
@@ -609,7 +609,6 @@ class Iota(StrictIota):
         """
 
         return extended.BroadcastBundleCommand(self.adapter)(tail_hash=tail_transaction_hash)
-
 
     def find_transaction_objects(
             self,
--- a/iota/api.py
+++ b/iota/api.py
@@ -585,6 +585,32 @@ class Iota(StrictIota):
         """
         return extended.BroadcastAndStoreCommand(self.adapter)(trytes=trytes)
 
+    def broadcast_bundle(self, tail_transaction_hash):
+        # type (TransactionHash) -> dict
+        """
+        Re-broadcasts all transactions in a bundle given the tail transaction hash.
+        It might be useful when transactions did not properly propagate,
+        particularly in the case of large bundles.
+
+        :param tail_transaction_hash:
+            Tail transaction hash of the bundle.
+
+        :return:
+            Dict with the following structure:::
+
+                {
+                    'trytes': List[TransactionTrytes],
+                        List of TransactionTrytes that were broadcast.
+                }
+
+        References:
+
+        - https://github.com/iotaledger/iota.js/blob/next/api_reference.md#module_core.broadcastBundle
+        """
+
+        return extended.BroadcastBundleCommand(self.adapter)(tail_hash=tail_transaction_hash)
+
+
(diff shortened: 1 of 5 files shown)


iota/api.py
```diff
@@ -585,6 +585,32 @@ def broadcast_and_store(self, trytes):
...
+        """
+        Re-broadcasts all transactions in a bundle given the tail transaction hash.
+        It might be useful when transactions did not properly propagate,
+        particularly in the case of large bundles.
+
+        :param tail_transaction_hash:
+            Tail transaction hash of the bundle.
+
+        :return:
+            Dict with the following structure:::
```
todofixthis:
```suggestion
            Dict with the following structure::
```

iota/commands/extended/broadcast_bundle.py
```diff
@@ -0,0 +1,51 @@
...
+        # and validates it.
+        # Returns List[List[TransactionTrytes]]
+        # (outer list has one item in current implementation)
+        bundle = GetBundlesCommand(self.adapter)(transaction=request['tail_hash'])
+        BroadcastTransactionsCommand(self.adapter)(trytes=bundle[0])
+        return {
+            'trytes': bundle[0],
+        }
+        # Future: Support multiple bundles in getBundlesCommand, and
+        # then broadcastBundle can do that as well.
```
todofixthis: ❔  This comment will probably get lost pretty quickly; could we create a GitHub issue for it, so that we don't lose track of it?

test/commands/extended/broadcast_bundle_test.py
```diff
@@ -0,0 +1,205 @@
...
+          self.trytes
+        )
+
+  def test_happy_path_multiple_bundle(self):
+    """
+    Test if command returns the correct bundle if underlying `get_bundles`
+    returns multiple bundles.
+    """
+    # Call the command with a tail hash.
+    # Lets mock away GetBundlesCommand, and we don't do
```
todofixthis:
```suggestion
    # Let's mock away GetBundlesCommand, and we don't do
```

test/commands/extended/broadcast_bundle_test.py
```diff
@@ -0,0 +1,205 @@
...
+    # BroadcastTransactionsCommand either.
+    # Note that GetBundlesCommand returns multiple bundles!
+    with patch('iota.commands.extended.get_bundles.GetBundlesCommand.__call__',
+               MagicMock(return_value=[self.trytes, self.trytes_dummy])
+              ) as mocked_get_bundles:
+      with patch('iota.commands.core.BroadcastTransactionsCommand.__call__',
+                 MagicMock(return_value= [])) as mocked_broadcast:
```
todofixthis:
```suggestion
                 MagicMock(return_value=[])) as mocked_broadcast:
```

iota/commands/extended/broadcast_bundle.py
lzpap: https://github.com/iotaledger/iota.py/issues/250

Issue #251, opened 2019-10-28
Make `is_promotable` helper function a proper extended API command


lzpap:
Current helper method only checks for consistency, but it should also verify that `attachmentTimestamp` is above a lower bound that is calculated based on number of milestones since the transaction attachment.

Reference from [iota.js](https://github.com/iotaledger/iota.js/blob/next/api_reference.md#module_core.isPromotable).

Make a proper extended API command out of the helper method.

PR #255, opened 2019-10-30
Implement is_promotable Extended API Command

lzpap:
Solves #251 

Checks if a list of tail transaction hashes can be promoted or not by calling `CheckConsistencyCommand` core API and performing extra checks on the `attachment_timestamp` of the transactions fetched from the Tangle with `GetTrytesCommand`.

If **all** tails are promotable, the returned `dict` contains only a `promotable` key and `True` value.
If **any** of the tails are not promotable, `promotable` is `False` and the `info` key contains information about what went wrong.

### DIFF
--- a/iota/commands/extended/is_promotable.py
+++ b/iota/commands/extended/is_promotable.py
@@ -81,8 +81,9 @@ class IsPromotableCommand(FilterCommand):
         }
 
         # Check timestamps
+        now = get_current_ms()
         for tx in transactions:
-            is_within = is_within_depth(tx.attachment_timestamp)
+            is_within = is_within_depth(tx.attachment_timestamp, now)
             if not is_within:
                 # Inform the user about what went wrong.
                 response['info'].append('Transaction {tx_hash} is above max depth.'.format(
@@ -93,7 +94,7 @@ class IsPromotableCommand(FilterCommand):
 
         # If there are no problems, we don't need 'info' field
         if response['promotable']:
-            del response['info']
+            response['info'] = None
 
         return response
 
@@ -106,10 +107,10 @@ class IsPromotableRequestFilter(RequestFilter):
                 f.FilterRepeater(f.Required | Trytes(TransactionHash)),
         })
 
-def is_within_depth(attachment_timestamp, depth=DEPTH):
+def is_within_depth(attachment_timestamp, now, depth=DEPTH):
+    # type (int, int, Optiona(int)) -> bool
     """
     Checks if `attachment_timestamp` is within limits of `depth`.
     """
-    now = get_current_ms()
     return attachment_timestamp < now and \
         now - attachment_timestamp < depth * MILESTONE_INTERVAL - ONE_WAY_DELAY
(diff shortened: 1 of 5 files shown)


iota/commands/extended/is_promotable.py
```diff
@@ -0,0 +1,115 @@
...
+                # Inform the user about what went wrong.
+                response['info'].append('Transaction {tx_hash} is above max depth.'.format(
+                    tx_hash=tx.hash
+                ))
+            # If one tx fails, response is false
+            response['promotable'] = response['promotable'] and is_within
+
+        # If there are no problems, we don't need 'info' field
+        if response['promotable']:
+            del response['info']
```
todofixthis: ❔ Could we keep it, but make it `None`?  That way the response always has consistent structure.

iota/commands/extended/is_promotable.py
```diff
@@ -0,0 +1,115 @@
...
+                f.Required |
+                f.Array |
+                f.FilterRepeater(f.Required | Trytes(TransactionHash)),
+        })
+
+def is_within_depth(attachment_timestamp, depth=DEPTH):
+    """
+    Checks if `attachment_timestamp` is within limits of `depth`.
+    """
+    now = get_current_ms()
```
todofixthis: ❔ Should this be cached for the duration of the command, so that each transaction in the bundle is checked against the same timestamp?

test/commands/extended/is_promotable_test.py
```diff
@@ -0,0 +1,407 @@
...
+
+    def test_happy_path(self):
+        """
+        Successfully checking promotability.
+        """
+        # To pass timestamp check, we need a timestamp that defines
+        # a moment no earlier than 660s from the moment the function
+        # is called during test. To do this, we cheat a bit:
+        tx = Transaction.from_tryte_string(self.trytes1)
+        tx.attachment_timestamp = get_current_ms()
```
todofixthis: ❔ Should we mock `get_current_ms` so that we can more easily control the outcome?

test/commands/extended/is_promotable_test.py
```diff
@@ -0,0 +1,407 @@
...
+                )],
+            }
+        )
+
+    def test_is_within_depth(self):
+        """
+        Test ``is_within_depth`` helper method.
+        """
+        # Timestamp is too old (depth=3)
+        old_timestamp = get_current_ms() - 660000
```
todofixthis: ❔ Should there be a reference to `MILESTONE_INTERVAL` here, instead of a magic number?

iota/commands/extended/is_promotable.py
lzpap: yes, agreed.

iota/commands/extended/is_promotable.py
lzpap: good catch, I will move this outside from here, so that it is only computed once before `# Check timestamps` part begin.

test/commands/extended/is_promotable_test.py
lzpap: yep

test/commands/extended/is_promotable_test.py
lzpap: in this method yes. But in `test_one_timestamp_invalid` it causes more headache, because then the 'modified' transaction becomes invalid and will not have a valid hash that can be compared to `self.hash2`

PR #261, opened 2019-11-12
Improve API documentation

lzpap:
## Changes
- Restructured API documentation into 3 parts:
  - API classes
  - Core API commands
  - Extended API commands

- Improve docstrings of API calls and types.
- Introduce autodoc support for sphinx to generate
  documentation from docstrings.
- Update Discord channel information

## View changes in browser

```
cd iota.py
pip install .[docs-builder]
cd docs
make html
```

Then you can open `docs/_build/html/index.html` to view the documentation.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -26,9 +26,9 @@ Dependencies
 ============
 PyOTA is compatible with Python 3.7, 3.6, 3.5 and 2.7
 
-============
-Installation
-============
+=============
+Install PyOTA
+=============
 To install the latest version::
 
   pip install pyota
--- a/iota/api.py
+++ b/iota/api.py
@@ -64,17 +64,45 @@ class StrictIota(object):
     References:
 
     - https://docs.iota.org/docs/node-software/0.1/iri/references/api-reference
+
+    :param AdapterSpec adapter:
+        URI string or BaseAdapter instance.
+
+    :param Optional[bool] testnet:
+        Whether to use testnet settings for this instance.
+        On the testnet, minimum weight magnitude is set to 9, on mainnet
+        it is 1 by default.
+
+    :param Optional[bool] local_pow:
+        Whether to perform proof-of-work locally by redirecting all calls
+        to :py:meth:`attach_to_tangle` to
+        `ccurl pow interface <https://pypi.org/project/PyOTA-PoW/>`_.
+
+            See :ref:`Optional Local Pow` for more info and
+            :ref:`find out<pow-label>` how to use it.
+
     """
     commands = discover_commands('iota.commands.core')
 
     def __init__(self, adapter, testnet=False, local_pow=False):
         # type: (AdapterSpec, bool, bool) -> None
         """
-        :param adapter:
+        :param AdapterSpec adapter:
             URI string or BaseAdapter instance.
 
-        :param testnet:
+        :param bool testnet:
             Whether to use testnet settings for this instance.
+            On the testnet, minimum weight magnitude is set to 9, on mainnet
+            it is 1 by default.
+
+        :param Optional[bool] local_pow:
+            Whether to perform proof-of-work locally by redirecting all calls
+            to :py:meth:`attach_to_tangle` to
+            `ccurl pow interface <https://pypi.org/project/PyOTA-PoW/>`_.
+
+                See :ref:`Optional Local Pow` for more info and
+                :ref:`find out<pow-label>` how to use it.
+
         """
         super(StrictIota, self).__init__()
 
@@ -106,7 +134,7 @@ class StrictIota(object):
         If you want to execute an arbitrary API command, use
         :py:meth:`create_command`.
 
-        :param command:
+        :param Text command:
             The name of the command to create.
 
         References:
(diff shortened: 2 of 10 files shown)


iota/api.py
```diff
@@ -472,6 +691,22 @@ def get_trytes(self, hashes):
...
+                        List of transaction trytes for the given transaction
+                        hashes (in the same order as the parameters).
+                    'duration': int,
+                        Number of milliseconds it took to complete the request.
+                }
+
+        .. note::
+            If a node doesn't have the trytes for a given transaction hash in
+            its ledger, the value at the index of that transaction hash is either
+            ``null`` or a string of 9s.
```
todofixthis: ❔ Should these lines be indented more, so that they're part of the `:return:` block?

iota/api.py
```diff
@@ -526,9 +792,23 @@ def were_addresses_spent_from(self, addresses):
...
             List of addresses to check.
 
+        :return:
+            ``dict`` with the following structure::
+
+                {
+                    'states': List[bool],
+                        States of the specified addresses in the same order as
+                        the values in the addresses parameter. A true value means
+                        that the address has been spent from..
```
todofixthis:
```suggestion
                        that the address has been spent from.
```

iota/api.py
```diff
@@ -526,9 +792,23 @@ def were_addresses_spent_from(self, addresses):
...
+        :param Iterable[Address] addresses:
             List of addresses to check.
 
+        :return:
+            ``dict`` with the following structure::
+
+                {
+                    'states': List[bool],
+                        States of the specified addresses in the same order as
+                        the values in the addresses parameter. A true value means
```
todofixthis:
```suggestion
                        the values in the addresses parameter. A ``true`` value means
```

iota/api.py
```diff
@@ -543,6 +823,25 @@ class Iota(StrictIota):
...
+        Seed used to generate new addresses.
+        If not provided, a random one will be generated.
+
+        .. note::
+            This value is never transferred to the node/network.
+
+    :param Optional[bool] testnet:
+        Whether to use testnet settings for this instance.
+        On the testnet, minimum weight magnitude is set to 9, on mainnet
+        it is 1 by default.
```
todofixthis: 💭 Wondering if maybe we should make this more generic, e.g., "On the testnet, minimum weight magnitude is decreased." and/or include a link to some documentation about IOTA testnet?  I feel like this info could get overlooked and end up out-of-date eventually.

iota/api.py
```diff
@@ -543,6 +823,25 @@ class Iota(StrictIota):
...
+
+    :param Optional[bool] local_pow:
+        Wether to perform proof-of-work locally by redirecting all calls
+        to :py:meth:`attach_to_tangle` to ccurl pow interface.
```
todofixthis:
```suggestion
        to :py:meth:`attach_to_tangle` to ccurl pow interface.
        
        See https://pyota.readthedocs.io/en/latest/#optional-local-pow for more info.
```

iota/api.py
```diff
@@ -873,45 +1186,45 @@ def get_new_addresses(
             security_level=AddressGenerator.DEFAULT_SECURITY_LEVEL,
             checksum=False,
     ):
-        # type: (int, Optional[int], int, bool) -> dict
+        # type: ( int, int, int, bool) -> dict
```
todofixthis:
```suggestion
        # type: (int, int, int, bool) -> dict
```

iota/api.py
```diff
@@ -252,18 +323,17 @@ def check_consistency(self, tails):
...
 
                 {
                     'state': bool,
-
-                    'info': str,
-                        This field will only exist set if ``state`` is
-                        ``False``.
+                        Wether tails resolve to consistent ledger.
+                    'info': Text,
+                        This field will only exist if 'state' is False.
```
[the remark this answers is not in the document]
lzpap: We can't control the response from the node in ``check_consistency``, so to make it consistent, I suggest we implement the same behavior in ``is_promotable``, that is to do
```
if response['promotable']:
    del response['info']
```
instead of the current ``response['info']=None`` in https://github.com/iotaledger/iota.py/blob/a68254d040dcf9a581a851f5953e056ff538897f/iota/commands/extended/is_promotable.py#L96

iota/api.py
lzpap: Good point!

iota/api.py
lzpap: Agreed, also created an issue to unify naming: #263

README.rst
```diff
@@ -26,9 +26,9 @@ Dependencies
 ============
 PyOTA is compatible with Python 3.7, 3.6, 3.5 and 2.7
 
-============
-Installation
-============
+=============
+Install PyOTA
```
lzpap: This change is needed to make all section titles unique due to `autosectionlabel` sphinx extension. There is a section with the same title in `getting_started.rst`.

todofixthis: 💭 Hmm... wondering if we should configure Travis build so that it also tries to build the docs, so that we can catch these automatically.

PR #265, opened 2019-11-15
Support adding/removing address checksums

lzpap:
## Changes
Two new methods in ``Address`` class:
- ``add_checksum``: Appends a valid checksum to the ``Address`` object. ``Address.checksum`` attribute is updated and checksum trytes are appended to internal tryte buffer.
- ``remove_checksum``: Slices off the checksum from an ``Address`` object. ``Address.checksum`` attribute is updated (to value ``None``) and internal tryte buffer is modified to hold only the 81 original address trytes.

### DIFF
--- a/test/types_test.py
+++ b/test/types_test.py
@@ -1049,6 +1049,81 @@ class AddressTestCase(TestCase):
     self.assertEqual(checked.key_index, 42)
     self.assertEqual(checked.balance, 86)
 
+  def test_add_checksum(self):
+    """
+    Checksum is added to an address without it.
+    """
+    addy = Address(
+      trytes =
+        b'ZKIUDZXQYQAWSHPKSAATJXPAQZPGYCDCQDRSMWWCGQJNI'
+        b'PCOORMDRNREDUDKBMUYENYTFVUNEWDBAKXMV'
+    )
+
+    addy.add_checksum()
+
+    self.assertTrue(addy.is_checksum_valid())
+    self.assertTrue(len(addy) == Address.LEN + AddressChecksum.LEN)
+
+  def test_add_checksum_second_time(self):
+    """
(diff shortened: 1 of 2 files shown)


test/types_test.py
```diff
@@ -1049,6 +1049,39 @@ def test_with_checksum_attributes(self):
...
+    addy = Address(
+      trytes =
+        b'ZKIUDZXQYQAWSHPKSAATJXPAQZPGYCDCQDRSMWWCGQJNI'
+        b'PCOORMDRNREDUDKBMUYENYTFVUNEWDBAKXMV'
+    )
+
+    addy.add_checksum()
+
+    self.assertTrue(addy.is_checksum_valid())
+    self.assertTrue(len(addy) == Address.LEN + AddressChecksum.LEN)
```
todofixthis: ❔ Could we also add a test for trying to add a checksum to an address that already has it?

test/types_test.py
```diff
@@ -1049,6 +1049,39 @@ def test_with_checksum_attributes(self):
...
+        b'PCOORMDRNREDUDKBMUYENYTFVUNEWDBAKXMVJJJGBARPB'
+    )
+
+    self.assertTrue(addy.is_checksum_valid())
+    self.assertTrue(len(addy) == Address.LEN + AddressChecksum.LEN)
+
+    addy.remove_checksum()
+
+    self.assertFalse(addy.is_checksum_valid())
+    self.assertTrue(len(addy) == Address.LEN)
```
todofixthis: ❔ Could we also add a test for trying to remove a checksum from an address that doesn't already have one?

test/types_test.py
lzpap: good idea!

PR #283, opened 2019-12-16
Add zero-value transaction example

JakeSCahill:


todofixthis:
Thanks @JakeSCahill !  Should we also add this example to `docs/README.rst`?

README.md
```diff
@@ -112,7 +112,40 @@ print(response)
...
 ## Examples
 
-You can find an example project in the [examples](https://github.com/iotaledger/iota.py/tree/master/examples) directory.
+We have a list of test cases in the [`examples` directory](https://github.com/iotaledger/iota.py/tree/master/examples) that you can use as a reference when developing apps with IOTA.
+
+Here's how you could send a zero-value transaction, using the library. For the guide, see the [documentation portal](https://docs.iota.org/docs/client-libraries/0.1/how-to-guides/python/send-your-first-bundle).
+
+```python
+# You don't need a seed to send zero-value transactions
+api = Iota('https://nodes.devnet.iota.org:443', testnet = True)
```
todofixthis:
```suggestion
api = Iota('https://nodes.devnet.iota.org:443', testnet=True)
```

From [PEP-8](https://www.python.org/dev/peps/pep-0008/#other-recommendations):
> Don't use spaces around the = sign when used to indicate a keyword argument, or when used to indicate a default value for an unannotated function parameter.

README.md
```diff
@@ -112,7 +112,40 @@ print(response)
...
+# that sends the message to the address
+tx = ProposedTransaction(
+    address = Address(address),
+    message = message,
+    value = 0
+)
+
+# Create a bundle from the `ProposedTransaction` object
+# and send the transaction to the node
+result = api.send_transfer(transfers = [tx])
```
todofixthis:
```suggestion
result = api.send_transfer(transfers=[tx])
```

README.md
```diff
@@ -112,7 +112,40 @@ print(response)
...
+We have a list of test cases in the [`examples` directory](https://github.com/iotaledger/iota.py/tree/master/examples) that you can use as a reference when developing apps with IOTA.
+
+Here's how you could send a zero-value transaction, using the library. For the guide, see the [documentation portal](https://docs.iota.org/docs/client-libraries/0.1/how-to-guides/python/send-your-first-bundle).
+
+```python
+# You don't need a seed to send zero-value transactions
+api = Iota('https://nodes.devnet.iota.org:443', testnet = True)
+
+# Define a message to send.
+# This message must include only ASCII characters.
```
todofixthis: 💭 PyOTA does support any Unicode code points; it encodes the value using UTF-8 before translating into trytes.

JakeSCahill: I thought we weren't recommending UTF-8 characters at the moment because this method isn't standardized across the libraries and may change.

PR #287, opened 2020-01-14
Add first tutorials to documentation

lzpap:
## Description
Open a new section `Tutorials`  in PyOTA documentation that will be home to detailed examples of tasks that could be carried out with the help of the library.

Proposed structure of the section:
 - Full code example first,
 - followed by a detailed, step-by-step code walk through.

Also, refer to other sections in the documentation while describing your code to help others develop their knowledge. If you notice that something is missing, improve it. (like this PR does for `TryteString.random()`.)

Added two basic scenarios that might be helpful to IOTA and Python newcomers:
 - `Hello Node` and
 - `Send Data`

### DIFF
--- a/docs/tutorials.rst
+++ b/docs/tutorials.rst
@@ -11,15 +11,16 @@ and suggestions in the `PyOTA Bug Tracker`_.
 
 Let's get to it then!
 
-Hello World
------------
+.. py:currentmodule:: iota
+
+Hello Node
+----------
 In this example, you will learn how to:
 
-- Import the ``iota`` package into your application.
-- Instantiate an API object for communication with the IOTA network.
-- Request information about the IOTA node you are connected to.
+- **Import the** ``iota`` **package into your application.**
+- **Instantiate an API object for communication with the IOTA network.**
+- **Request information about the IOTA node you are connected to.**
 
-.. py:currentmodule:: iota
 
 Code
 ~~~~
@@ -43,7 +44,8 @@ Code
 
 Discussion
 ~~~~~~~~~~
-::
+.. code-block::
+    :lineno-start: 1
 
     # Import neccessary modules
     from iota import Iota
@@ -60,7 +62,8 @@ the network. Read more about API objects at :ref:`PyOTA API Classes`.
 We also import the ``pprint`` method that prettifies the output before printing
 it to the output.
 
-::
+.. code-block::
+    :lineno-start: 5
 
     # Declare an API object
     api = Iota('https://nodes.devnet.iota.org:443')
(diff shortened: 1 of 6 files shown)


docs/tutorials.rst
```diff
@@ -0,0 +1,211 @@
...
+
+.. code-block::
+    :lineno-start: 13
+
+    # Tag is optional here
+    my_tag = Tag(b'MY9FIRST9TAG')
+
+To tag our transaction, we might define a custom :py:class:`Tag` object. Notice,
+that the ``b`` means we are creating a bytestring now. Each byte in it is
+interpreted as a tryte, therefore we are restricted to the tryte aphabet.
```
```rst
To tag our transaction, we might define a custom :py:class:`Tag` object. Notice,
that the ``b`` means we are passing a `bytestring`_ value instead of a unicode
string. This is so that PyOTA interprets our input as literal trytes, rather than a
unicode string that needs to be encoded into trytes.

When passing a bytestring to a PyOTA class, each byte is interpreted as a tryte;
therefore we are restricted to the `tryte aphabet`_.

.. _bytestring: https://docs.python.org/3/library/stdtypes.html#bytes
.. _tryte alphabet: ???
```

docs/tutorials.rst
```diff
@@ -0,0 +1,211 @@
...
+    from pprint import pprint
+
+    # Declare an API object
+    api = Iota('https://nodes.devnet.iota.org:443')
+
+    # Request information about the node
+    response = api.get_node_info()
+
+    # Using pprint instead of print for a nicer looking result in the console
+    pprint(response)
```
```rst
 .. literalinclude:: ../examples/tutorials/01_hello_node.py
    :linenos:
```
?

docs/tutorials.rst
```diff
@@ -0,0 +1,211 @@
...
+
+- **Encode data to be stored on the Tangle.**
+- **Generate a random IOTA address that doesn't belong to anyone.**
+- **Create a zero-value transaction with custom payload.**
+- **Send a transaction to the network.**
+
+Code
+~~~~
+.. literalinclude:: ../examples/tutorials/02_send_data.py
+   :linenos:
```
todofixthis: Nice!

docs/tutorials.rst
lzpap: good point, forgot it here :smiley:

docs/tutorials.rst
```diff
@@ -0,0 +1,211 @@
...
+
+Discussion
+~~~~~~~~~~
+.. code-block::
+    :lineno-start: 1
```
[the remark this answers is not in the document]
lzpap: yes, makes sense

docs/tutorials.rst
lzpap: well said! :)

PR #289, opened 2020-01-15
docs: Add 'Fetch Data' tutorial

lzpap:
## Description
- Add `Fetch Data` tutorial to gather messages from the Tangle.
- Introduce numbering in headings for tutorials.

### DIFF
--- a/docs/tutorials.rst
+++ b/docs/tutorials.rst
@@ -13,8 +13,8 @@ Let's get to it then!
 
 .. py:currentmodule:: iota
 
-Hello Node
-----------
+1. Hello Node
+-------------
 In this example, you will learn how to:
 
 - **Import the** ``iota`` **package into your application.**
@@ -73,8 +73,8 @@ we could list the ``features`` of the node::
 
     pprint(response['features'])
 
-Send Data
----------
+2. Send Data
+------------
 In this example, you will learn how to:
 
 - **Encode data to be stored on the Tangle.**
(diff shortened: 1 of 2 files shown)


docs/tutorials.rst
```diff
@@ -167,6 +167,66 @@ Observe how we extract the transaction hash from the response ``dict``. We take
...
+.. note::
+
+    Remember, that the parameters need to be supplied as lists, even if
+    there is only one value.
+
+.. literalinclude:: ../examples/tutorials/03_fetch_data.py
+   :lines: 16-29
+   :lineno-start: 16
+
+Finally, we extract the data we our looking for from the transaction objects.
```
```suggestion
Finally, we extract the data we are looking for from the transaction objects.
```

examples/tutorials/03_fetch_data.py
```diff
@@ -0,0 +1,29 @@
...
+    print('Found:')
+    # Iterate over the fetched transaction objects
+    for tx in response['transactions']:
+        # data is in the signature_message_fragment attribute as trytes, we need
+        # to decode it into a unicode string
+        try:
+            data = tx.signature_message_fragment.decode()
+        # Ignore invalid or not unicode encoded data 
+        except (TrytesDecodeError, UnicodeDecodeError):
+            continue
```
todofixthis: I think we can achieve the same result with:

```python
data = tx.signature_message_fragment.decode(errors='ignore')
```

More info:
https://github.com/iotaledger/iota.py/blob/2.1.0/iota/types.py#L536-L547

(this is meant to mimic the way Python's [`codecs.decode()` function](https://docs.python.org/3/library/codecs.html#codecs.decode) works)

lzpap: You are right, however, with `ignore`, only the values that can't be decoded are dropped, the rest of the non-meaningful message will be printed.
Anyway, I added a small explanation and changed it to your suggestion.

PR #292, opened 2020-01-21
fix: traverse_bundle checks for zero trytes

lzpap:
## Description
`get_bundles` calls `traverseBundleCommand` internally, which in turn calls `getTrytesCommand` to fetch the trytes for transactions in the bundle.

> If a node doesn't have the trytes for a given transaction hash in its ledger, the value at the index of that transaction hash is either null or a string of 9s.

Therefore, `getTrytesCommand`  doesn't return an empty list, rather a list with an empty transaction in it, meaning that all trytes of the transaction are '9's.

Detecting such cases when trying to fetch a bundle from the Tangle and throwing an exception.

### DIFF
--- a/iota/commands/extended/traverse_bundle.py
+++ b/iota/commands/extended/traverse_bundle.py
@@ -7,7 +7,7 @@ from typing import List, Optional
 import filters as f
 
 from iota import BadApiResponse, BundleHash, Transaction, \
-    TransactionHash, TryteString, Bundle
+    TransactionHash, TryteString, Bundle, TransactionTrytes
 from iota.commands import FilterCommand, RequestFilter
 from iota.commands.core.get_trytes import GetTrytesCommand
 from iota.exceptions import with_context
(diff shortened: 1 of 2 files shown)


iota/commands/extended/traverse_bundle.py
```diff
@@ -55,10 +55,13 @@ def _traverse_bundle(self, txn_hash, target_bundle_hash):
...
 
-        if not trytes:
+        # If no tx was found by the node for txn_hash, it returns 9s,
+        # so we check here if it returned all 9s trytes.
+        if not trytes or trytes == [TransactionTrytes('')]:
             raise with_context(
                 exc=BadApiResponse(
-                    'Bundle transactions not visible '
+                    'Could not get trytes of bundle transaction from the Tangle. '
+                    'Bundle transactions not visible.'
```
todofixthis: 💭 Wondering if this also indicates an issue with the IRI node — should `getTrytes` ever return an all-9's value?

lzpap: According to the [IRI API reference](https://docs.iota.org/docs/node-software/0.1/iri/references/api-reference#gettrytes), it may return string of 9s when the node doesn't have the transaction in its ledger.

todofixthis: Ah, I see.   That makes sense, since the client can pass multiple hashes to `getTrytes`.

PR #293, opened 2020-01-23
`get_bundles` can return multiple bundles

lzpap:
## Changes
- `get_bundles` extended API call accepts a list of tail transaction hashes instead of just one.
- `get_bundles` returns multiple bundles in the same order as the input tail hashes.
- Added test coverage for returning multiple bundles.
- Modified calls to `GetBundlesCommand()` throughout the lib.

WARNING: Breaking change!

### DIFF
--- a/iota/commands/extended/get_bundles.py
+++ b/iota/commands/extended/get_bundles.py
@@ -8,8 +8,8 @@ from iota import BadApiResponse, TransactionHash
 from iota.commands import FilterCommand, RequestFilter
 from iota.commands.extended.traverse_bundle import TraverseBundleCommand
 from iota.exceptions import with_context
-from iota.filters import Trytes
 from iota.transaction.validator import BundleValidator
+from iota.filters import Trytes
 
 __all__ = [
     'GetBundlesCommand',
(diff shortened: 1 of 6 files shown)


iota/commands/extended/get_bundles.py
```diff
@@ -31,35 +31,41 @@ def get_response_filter(self):
...
+                        'errors': validator.errors,
+                    },
+                )
+
+            bundles.append(bundle)
 
         return {
             # Always return a list, so that we have the necessary
             # structure to return multiple bundles in a future
             # iteration.
```
todofixthis: 💭 I'd say this comment can be removed now 😺

iota/commands/extended/get_bundles.py
```diff
@@ -31,35 +31,41 @@ def get_response_filter(self):
...
-            'bundles': [bundle],
+            'bundles': bundles,
         }
 
 class GetBundlesRequestFilter(RequestFilter):
     def __init__(self):
         super(GetBundlesRequestFilter, self).__init__({
-            'transaction': f.Required | Trytes(TransactionHash),
+            'transactions': f.Required | StringifiedTrytesArray(TransactionHash)
```
todofixthis:
```suggestion
            'transactions': f.Required | f.Array | f.FilterRepeater(f.Required | Trytes(TransactionHash))
```

❕ `StringifiedTrytesArray` converts the values into strings.  See `iota.commands.extended.send_trytes.SendTrytesRequestFilter` for a related example.

test/commands/extended/get_bundles_test.py
```diff
@@ -23,18 +23,24 @@ def setUp(self):
...
-        )
+        self.transactions = [
+            (
+                'TESTVALUE9DONTUSEINPRODUCTION99999KPZOTR'
+                'VDB9GZDJGZSSDCBIX9QOK9PAV9RMDBGDXLDTIZTWQ'
+            ),
+            (
+                'TESTVALUE9DONTUSEINPRODUCTION99999KPZOTR'
+                'VDB9GZDJGZSSDCBIX9QOK9PAV9RMDBGDXLDTIZTWQ'
+            ),
```
todofixthis: ❔ Should these values be different, to more accurately simulate real-world usage?

test/commands/extended/get_bundles_test.py
```diff
@@ -92,29 +103,43 @@ def test_fail_unexpected_parameters(self):
...
+        """
+        ``transactions`` has one hash rather than a list of hashes.
+        """
+        self.assertFilterErrors(
+            {
+                'transactions': self.transactions[0],
+            },
+
+            {
+                'transactions': [f.Type.CODE_WRONG_TYPE],
```
todofixthis: ❔ Could we also add tests for invalid values inside of `transactions`?  See `test.commands.extended.send_trytes_test.SendTrytesRequestFilterTestCase.test_fail_trytes_contents_invalid` for an example.

test/commands/extended/get_bundles_test.py
```diff
@@ -308,15 +333,45 @@ def test_happy_path(self):
             'trytes': [self.spam_trytes],
         })
 
-        response = self.command(transaction = self.tx_hash)
+        response = self.command(transactions = [self.tx_hash])
+
+        self.maxDiff = None
```
todofixthis: 💭 I see that we do this in a few places (and I am including me in that "we" 😅); it just occurred to me that this change doesn't get reverted after each test.  Is that something we should fix?

iota/commands/extended/get_bundles.py
lzpap: I see, going to replace it.

test/commands/extended/get_bundles_test.py
lzpap: What behavior do we exactly want? ``maxDiff=None`` disables the limit on diff output, does it affect the actual test result, or does it only makes the output visible when it fails? Could you elaborate a bit more on what should be fixed? :smiley:

test/commands/extended/get_bundles_test.py
lzpap: agreed

test/commands/extended/get_bundles_test.py
todofixthis: Yeah, that's true.  I guess it's just the purist in me — modifying a variable outside the test scope and not restoring the value after the test finishes? 🙀 Blaphemy! 😅

Okiedoke; no need to change this 😸

test/commands/extended/get_bundles_test.py
```diff
@@ -92,29 +103,73 @@ def test_fail_unexpected_parameters(self):
...
+            },
+
+            {
+                'transactions': [f.Type.CODE_WRONG_TYPE],
+            },
+        )
+
+    def test_fail_transactions_contents_invalid(self):
+        """
+        ``transactions`` is a non-empyt array, but it contains invlaid values.
```
todofixthis:
```suggestion
        ``transactions`` is a non-empty array, but it contains invalid values.
```

PR #297, opened 2020-01-29
Fixes #288 Remove tips from `get_inclusion_states`

### LINKED ISSUE #288: Remove tips parameter from get_inclusion_states API call
## Description
With the next IRI release v1.8.6, `tips` parameter will be removed from `getInclusionStates` API endpoint. See iotaledger/iri#1350 and iotaledger/iri#1685 for more info.

> IRI will only check if the tx was approved by a milestone. If clients still send the parameter it should be ignored.

## Changes needed
 - Remove the parameter in PyOTA from `get_inclusion_states()` and `getInclusionStatesCommand`.
- Update test case for the command.
- Update the documentation.

lzpap:
### Description
- `tips` parameter for ``get_inclusion_states`` is no longer in use from IRI 1.8.6.
- `GetInclusionStates` API endpoint returns, if tx is confirmed since the last milestone.
- `get_latest_inclusion_states` extended API becomes irrelevant, remove in this commit.
- Modify components using it, rely on `get_inclusion_states`.
- Update tests.
- Update documentation.


test/commands/core/get_inclusion_states_test.py
```diff
@@ -116,12 +87,12 @@ def test_fail_unexpected_parameters(self):
...
 
-        # I bring scientists, you bring a rock star.
-        'foo': 'bar',
+        # 'tips' deprecated in IRI 1.9.0
+        'tips': [self.trytes1]
       },
 
       {
-        'foo': [f.FilterMapper.CODE_EXTRA_KEY],
+        'tips': [f.FilterMapper.CODE_EXTRA_KEY],
```
todofixthis: 👍🏻 Good thinking!

PR #298, opened 2020-01-29
Fixes #263 Rename `testnet` to `devnet`

### LINKED ISSUE #263: Rename testnet to devnet
PyOTA calls the network used for development ``testnet``, however it should be called ``devnet``, as in the official [docs](https://docs.iota.org/docs/getting-started/0.1/network/iota-networks/).

lzpap:


docs/tutorials.rst
```diff
@@ -100,8 +100,8 @@ Discussion
 We have seen this part before. Note, that now we import more objects which we
 will use to construct our transaction.
 
-Notice ``testnet=True`` in the argument list of the API instantiation. We
-tell the API directly that we will use the devnet/testnet. By default, the API
+Notice ``devnet=True`` in the argument list of the API instantiation. We
+tell the API directly that we will use the devnet. By default, the API
```
```suggestion
tell the API directly that we will use IOTA's testnet, known as the devnet. By default, the API
```

❔ I don't think that devnet is a commonly-used term outside of IOTA.  For new developers, I think it may be useful to include a bit of extra context.

lzpap: agreed. With this renaming, we try to establish a common naming within IOTA at least :smile:

PR #301, opened 2020-02-11
Asynchronous API, Networking and Asyncio Support

lzpap:
### Motivation
So far, PyOTA has been relying on synchronous communication with nodes. This meant, that API requests could only be sent sequentially, one after the other. While this is simple and easy for developers to comprehend, it doesn't scale well and wastes a lot of precious time waiting for the results of network IO bound tasks.

### Solution
With this PR, PyOTA is capable of doing asynchronous networking. Requests can be sent concurrently with the help of `asyncio` and `httpx` packages. When using the new `AsyncIota` API class, one can schedule API calls to execute concurrently. The API calls in the new class are actually [python coroutines](https://docs.python.org/3/library/asyncio-task.html), but both their argument lists and return values are preserved from the previous version of PyOTA.

Regular, synchronous API calls are still supported with the "old" `Iota` API class. There should be no difference from a user perspective to issue a synchronous API call, however, the methods have been reworked to execute the aforementioned coroutines inside an event loop under the hood. This speeds up some extended API calls too.

### Changes in source code
- Use `AsyncClient` from `httpx` for networking instead of `requests`. `httpx` supports Python 3.6+.
- Mimic the behavior of `AsyncClient` in `MockAdapter`, return an `asyncio.Future` response object for a request.
- Refactor `RoutingWrapper` to work with async.
- Create new API classes:
   - `AsyncStrictIota` and
   - `AsyncIota`.
  These API classes define the API calls as coroutines rather than methods. Hence, the API calls can be awaited by the application developer.
- Refactor `StrictIota` and `Iota` classes to build on top of the two new classes. They provide the synchronous method interface by executing the parents' coroutines in an event loop and returning their result.
- Refactor commands to work asynchronously. Modify `BaseCommand.__call__()` and `BaseCommand._execute()` to work as coroutines.
- All core commands are async.
- Update extended commands to work as async.
  - `BroadcastAndStoreCommand` is (nearly) twice as fast as before, which speeds up for example `SendTrytesCommand` and `SendTransferCommand` as well.
  - `GetBundlesCommand` is faster when called with multiple args.
  - `GetNewAddressesCommand` is faster as internal address checks are done concurrently.
- Remove `Helpers` class as `IsPromotableCommand` is properly implemented as an extended API call.
- Update `iota.multisig` module to work async, create new `AsyncMultisigIota` API class and rebase the original multisig api class on top of it, using the same logic as above.

- **Remove Support for Pyhton 2 and Python 3.5!**
  Async modules do not work well with `3.6>`.

### Changes in test code
- Make use of `aiounittest` module for testing async coroutines. `@async_test` decorator runs a test method inside an event loop.
- Update adapter and api tests.
- Update core API command tests for async.
- Update extended command tests for async.
- Update multisig tests for async.
- Remove PY2 and PY3.5 scenarios from test configuration.

### Future work in another PR(s):
- Update documentation with new API classes and methods.
- Write examples to show how to use the new async classes.
- Clean up PY2 code snippets from codebase.
- Transform the codebase to PY3 format.


iota/commands/extended/utils.py
```diff
@@ -19,7 +19,8 @@
 from iota.crypto.types import Seed
 
 
-def iter_used_addresses(
+# This is an async generator!
+async def iter_used_addresses(
```
todofixthis:
```py
 async def iter_used_addresses(
    """
    .. important::
        This is an async generator!
    """
```

setup.py
```diff
@@ -30,6 +30,7 @@
 tests_require = [
     'mock; python_version < "3.0"',
     'nose',
+    'aiounittest',
 ]
```
todofixthis: There goes all my hard work keeping imports, `__all__`, etc. in alphabetical order... 😹

test/commands/extended/get_new_addresses_test.py
```diff
@@ -472,18 +503,22 @@ def test_get_addresses_online_already_spent_from(self):
     # it skipped that one.
     self.assertDictEqual(response, {'addresses': [self.addy_2]})
 
-    self.assertListEqual(
+    # Due to running WereAddressesSpentFromCommand and FindTransactionsCommand
+    # with asyncio.gather, we can't infer their execution order. Therefore,
+    # we need to assert if the contents of the two lists match by value,
+    # regardless of their order:
+    # https://docs.python.org/3.5/library/unittest.html#unittest.TestCase.assertCountEqual
+    self.assertCountEqual(
```
todofixthis: 👍🏻 Nice!

test/commands/extended/utils_test.py
```diff
@@ -239,3 +245,5 @@ def test_multiple_addresses_return(self):
                 },
             ]
         )
+
+# TODO: add tests for `get_bundles_from_transaction_hashes`
```
todofixthis: ❔ Should we capture this with a GitHub issue, so that we don't forget about it?

tox.ini
```diff
@@ -4,10 +4,11 @@
 # and then run "tox" from this directory.
 
 [tox]
-envlist = py27, py35, py36, py37
+envlist = py36, py37
```
todofixthis: ❔ Should we also update the list of supported Python versions in `README.md` and `docs`?

test/commands/extended/utils_test.py
lzpap: yep, see #302

tox.ini
lzpap: absolutely, I forgot about it.

iota/api.py
```diff
@@ -749,15 +751,293 @@ def were_addresses_spent_from(self, addresses):
...
+
+    def add_neighbors(self, uris):
+        # type: (Iterable[Text]) -> dict
+        # Copy parent's docstring
+        __doc__ = super(StrictIota, self).add_neighbors.__doc__
+
+        # Execute original coroutine inside an event loop to make this method
+        # synchronous
+        return asyncio.get_event_loop().run_until_complete(
+            super(StrictIota, self).add_neighbors(uris)
```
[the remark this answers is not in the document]
lzpap: agreed :+1:

iota/api.py
```diff
@@ -749,15 +751,293 @@ def were_addresses_spent_from(self, addresses):
 
         - https://docs.iota.org/docs/node-software/0.1/iri/references/api-reference#wereaddressesspentfrom
         """
-        return core.WereAddressesSpentFromCommand(self.adapter)(
+        return await core.WereAddressesSpentFromCommand(self.adapter)(
             addresses=addresses,
         )
 
+class StrictIota(AsyncStrictIota):
```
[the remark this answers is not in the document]
lzpap: Good idea! 
One question I have in mind is if the IDE will pick up this for code suggestions ( like intellisense)?
I really like it when I start typing a method name and it shows me the options, what arguments it takes and so on, basically the doctsrings of the method.
I don't want to loose the support for this, as it helps developers to navigate in the already complex world of iota :smiley:

iota/api.py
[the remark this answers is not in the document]
lzpap: btw, @todofixthis could you explain me why `super()` is called in a base class?
For example here:
```python
@add_metaclass(AdapterMeta)
class BaseAdapter(object):
    """
    Interface for IOTA API adapters.

    Adapters make it easy to customize the way an API instance
    communicates with a node.
    """
    supported_protocols = ()  # type: Tuple[Text]
    """
    Protocols that ``resolve_adapter`` can use to identify this adapter
    type.
    """

    def __init__(self):
        super(BaseAdapter, self).__init__()
```

setup.py
lzpap: hahaha sorry, didn't notice :smile_cat: 
gonna fix this!

test/commands/extended/utils_test.py
todofixthis: Oh, right 😅

iota/api.py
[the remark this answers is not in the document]
todofixthis: 🤔 Good question.  I think the reason is in case a derived class uses multiple inheritance, e.g., `CustomAdapter(BaseAdapter, OtherBase):` (or possibly the other way around; multiple inheritance is so confusing 😅) — without the super call, it's possible that the constructor for one or more bases won't get called.

iota/api.py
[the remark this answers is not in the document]
todofixthis: True that.  IDEs definitely won't pick up on it (I can attest to this because it was the same reason for dropping the command registry and making explicit API methods in the first place 😹).

iota/api.py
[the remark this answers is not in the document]
lzpap: Okay, thanks for the answer!

Let's say we switch to py3-style `super()` calls everywhere and don't need to explicitly tell `type` and `object` to `super()`. If I do it with `BaseAdapter`:
```python
@add_metaclass(AdapterMeta)
class BaseAdapter:
    """
    Interface for IOTA API adapters.

    Adapters make it easy to customize the way an API instance
    communicates with a node.
    """
    supported_protocols = ()  # type: Tuple[Text]
    """
    Protocols that ``resolve_adapter`` can use to identify this adapter
    type.
    """

    def __init__(self):
        super().__init__()
``` 
I get a `TypeError` in the `super()` call:
```
Traceback (most recent call last):
  File "/pyota/test/commands/extended/utils_test.py", line 16, in setUp
    self.adapter = MockAdapter()
  File "/pyota/iota/adapter/__init__.py", line 547, in __init__
    super(MockAdapter, self).__init__()
  File "/pyota/iota/adapter/__init__.py", line 172, in __init__
    super().__init__()
TypeError: super(type, obj): obj must be an instance or subtype of type
```
, but the MRO seems okay:
```
>>> from iota import BaseAdapter
>>> BaseAdapter.__mro__
(<class 'iota.adapter.BaseAdapter'>, <class 'object'>)
```
What am I missing here? :thinking:

iota/api.py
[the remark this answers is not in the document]
lzpap: Well, I guess then we have to swallow the inconvenience of maintaining methods/docstrings for the convenience of users of PyOTA :)

iota/api.py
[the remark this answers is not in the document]
todofixthis: Thanks for checking that out!  Hmmm.. maybe it's not needed in Python 3 then?  I don't recall the exact situation where I thought it was necessary; maybe we just remove it and see if anyone complains 😅

PR #306, opened 2020-02-20
Install requirements on RTD

lzpap:


docs/requirements.txt
```diff
@@ -0,0 +1 @@
+sphinxcontrib-asyncio
```
todofixthis: 👍🏻 Ah, cool; I didn't know that RTD could have its own `requirements.txt` file.

PR #316, opened 2020-03-04
docs: Add `Creating Transfers` section

lzpap:
## Description
- Explanation of how transfers are created
- Two new figures on transfer creation and available API methods.
- 3 code examples to show the different ways of transaction creation.

Other changes:
- Renamed `Transaction.get_transaction_validation_trytes()` to better reflect its purpose.
- Add file names to auto generated labels in sphinx.


docs/transfers.rst
```diff
@@ -0,0 +1,296 @@
+Creating transfers
+==================
+
+IOTA is a permissionless DLT solution, therefore anyone can send transactions
+to the network and initiate transfers. The IOTA client libraries help you to
+abstract away low level operations required to construct and send a transfer
```
todofixthis:
```suggestion
abstract away low-level operations required to construct and send a transfer
```

docs/transfers.rst
```diff
@@ -0,0 +1,296 @@
...
+
+.. note::
+
+    Before proceeding, make sure you read and understood the
+    :ref:`basic_concepts:Basic Concepts` and :ref:`types:PyOTA Types` sections!
+
+Anatomy of a Transfer
+---------------------
+
+We already now that the Tangle consists of :ref:`transactions <basic_concepts:Transaction>`
```
todofixthis:
```suggestion
We already know that the Tangle consists of :ref:`transactions <basic_concepts:Transaction>`
```

docs/transfers.rst
```diff
@@ -0,0 +1,296 @@
...
+`Zero-value bundles`_ contain only zero value transactions, while
+`transfer bundles`_ may also contain input and output transactions.
+
+But how to construct these bundles and send them to the network?
+
+The process can be boiled down to 5 steps:
+
+    1. Create individual transaction(s).
+    2. Construct a bundle from the transaction(s).
+    3. Obtain references to two transactions waiting to be confirmed (tips) from the Tangle.
```
todofixthis:
```suggestion
    3. Obtain references to two unconfirmed transactions ("tips") from the Tangle.
```

docs/transfers.rst
```diff
@@ -0,0 +1,296 @@
...
+
+.. py:currentmodule:: iota
+
+1. Create Transactions
+~~~~~~~~~~~~~~~~~~~~~~
+The first step is to create the individual transaction objects. You have to
+specify ``address`` and ``value`` for each transaction. A negative ``value``
+means spending from ``address``. Furthermore, you can define a ``tag``, and for
+zero-value transactions, a ``message``. ``timestamp`` is usually auto-generated
+by the IOTA libraries.
```
todofixthis:
```suggestion
The first step is to create the individual transaction objects. You have to
specify ``address`` and ``value`` for each transaction.  Furthermore, you can define a ``tag``, and for
zero-value transactions, a ``message``.  A ``timestamp`` is also required, though this value is usually auto-generated
by the IOTA libraries.

.. note::
  Unlike on other decentralised ledgers, IOTA transactions can have positive *or* negative ``value`` amounts.  In order to send iotas from one address to another, at least two transactions are required:
  
  * one with *positive* ``value`` (to increment the balance of the receiver), and
  * one with *negative* ``value`` (to decrement the balance of the sender).
```

docs/transfers.rst
```diff
@@ -0,0 +1,296 @@
...
+
+In PyOTA, use :py:class:`ProposedTransaction` to declare transactions.
+
+2. Create Bundle from Transactions
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+A bundle is a collection of transactions, treated as an atomic unit
+when sent to the network. A bundle makes a value (token) transfer possible by
+grouping together input an output transactions.
```
todofixthis:
```suggestion
grouping together input and output transactions.
```

docs/transfers.rst
```diff
@@ -0,0 +1,296 @@
...
+
+A bundle always has to be balanced: the sum of ``value`` attributes of the
+transactions in the bundle should always be zero. Transactions in the bundle
+are indexed individually, and also contain information on how many other
```
todofixthis:
```suggestion
are also indexed individually and contain information on how many other
```

docs/transfers.rst
```diff
@@ -0,0 +1,296 @@
...
+means spending from ``address``. Furthermore, you can define a ``tag``, and for
+zero-value transactions, a ``message``. ``timestamp`` is usually auto-generated
+by the IOTA libraries.
+
+In PyOTA, use :py:class:`ProposedTransaction` to declare transactions.
+
+2. Create Bundle from Transactions
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+A bundle is a collection of transactions, treated as an atomic unit
+when sent to the network. A bundle makes a value (token) transfer possible by
```
todofixthis:
```suggestion
when sent to the network. A bundle makes a value (iota token) transfer possible by
```

docs/transfers.rst
```diff
@@ -0,0 +1,296 @@
...
+transactions in the bundle should always be zero. Transactions in the bundle
+are indexed individually, and also contain information on how many other
+transactions there are in the bundle.
+
+Once complete, a bundle has to be finalized to generate the bundle hash based
+on the `bundle essence`_. The bundle hash is the unique identifier of the
+bundle.
+
+After finalization, input transactions in the bundle need to be signed to prove
+ownership of tokens being transferred.
```
todofixthis:
```suggestion
ownership of iotas being transferred.
```

docs/transfers.rst
```diff
@@ -0,0 +1,296 @@
...
+
+:py:class:`ProposedBundle` helps you in PyOTA to create bundles, add transactions,
+finalize the bundle and sign the inputs.
```
todofixthis:
```suggestion
.. tip:
  :py:class:`ProposedBundle` helps you in PyOTA to create bundles, add transactions,
  finalize the bundle and sign the inputs.  We'll see how to use :py:class:`ProposedBundle` in
  :ref:`Use the Library` below.
```

docs/transfers.rst
```diff
@@ -0,0 +1,296 @@
...
+
+3. Select two tips
+~~~~~~~~~~~~~~~~~~
+
+Tips are transactions that are yet to be confirmed by the network. We can
+obtain two tips by requesting them from a node. In PyOTA, :py:meth:`~Iota.get_transactions_to_approve`
+does the job: it returns a ``trunk`` and a ``branch`` :py:class:`TransactionHash`.
+
+Our bundle will validate these two transactions once in the Tangle.
```
todofixthis:
```suggestion
Because our bundle references these two transactions, it will validate them once it is added to the Tangle.
```

docs/transfers.rst
```diff
@@ -0,0 +1,296 @@
...
+
+4. Do Proof-of-Work
+~~~~~~~~~~~~~~~~~~~
+
+The bundle has been finalized, inputs have been signed, we have two tips,
```
todofixthis:
```suggestion
The bundle has been finalized, inputs have been signed, and we have two tips;
```

docs/transfers.rst
```diff
@@ -0,0 +1,296 @@
...
+now it's time to prepare the bundle to be attached to the Tangle. All
+transactions reference two other transactions in the Tangle, therefore we need
+to select these references for each transaction in our bundle.
```
todofixthis:
```suggestion
now it's time to prepare the bundle to be attached to the Tangle. As noted in the previous section, every
transaction references two other transactions in the Tangle; therefore we need
to select these references for each transaction in our bundle.
```

docs/transfers.rst
```diff
@@ -0,0 +1,296 @@
...
+   :scale: 100 %
+   :alt: Bundle structure with four transactions.
+
+   Structure of a bundle with four transactions. Numbers in brackets denote
+   (``currentIndex``, ``lastIndex``) fields. Head of the bundle has index 3,
+   while tail has index 0.
+
+For all non-head transactions in the bundle, the trunk reference is the next
+transaction in the bundle, while the branch reference is the trunk transaction
+hash, one of the tips.
```
todofixthis:
```suggestion
transaction in the bundle, while the branch reference is one of the tips.
```

I thought that branch is the same for all transactions in a bundle (including the head tx)?

docs/transfers.rst
```diff
@@ -0,0 +1,296 @@
...
+
+The head transaction is different: the trunk reference is the trunk tip, while
+the branch reference is the branch tip.
```
todofixthis:
```suggestion
The head transaction is a little bit different.  Its branch reference is the same tip as for the other transactions in the bundle, but since it is the last transaction in the bundle, its trunk reference is the other selected tip.
```

docs/transfers.rst
```diff
@@ -0,0 +1,296 @@
...
+The proof-of-work calculation has to be done for each transaction individually,
+therefore the more transactions you have in the bundle, the more time it will
+take. The difficulty of the calculation also depends on the `minimum weight magnitude`_
+set by the network.
+
+The output of the proof-of-work algorithm is a ``nonce`` value that is appended
+to the the transaction, resulting in the attached transaction trytes.
+Nodes validate the proof-of-work of a transaction by calculating the transaction's
+hash from the attached transaction trytes. If the resulting hash has at least
+``minimum weight magnitude`` number of trailing zero trits, it is correct.
```
todofixthis:
```suggestion
``minimum weight magnitude`` number of trailing zero trits, the transaction is valid.
```

docs/transfers.rst
```diff
@@ -0,0 +1,296 @@
...
+
+5. Broadcast and Store
+~~~~~~~~~~~~~~~~~~~~~~
+
+The final step is to send the bundle to the network. Nodes will broadcast
+the transactions in the network, and store them in their local database.
+
+In PyOTA, use :py:meth:`~Iota.broadcast_and_store` to achieve this.
+
+Observe the bird-eye view of the Tangle depicted at the last step of the
```
todofixthis:
```suggestion
Observe the bird's-eye view of the Tangle depicted at the last step of the
```

docs/transfers.rst
```diff
@@ -0,0 +1,296 @@
...
+process. Our transactions are part of the Tangle, reference each other and
```
todofixthis:
```suggestion
process. Our transactions are part of the Tangle, referencing each other and
```

docs/transfers.rst
```diff
@@ -0,0 +1,296 @@
...
+the two tips. Newer transactions may reference our transactions as branch or
+trunk.
```
todofixthis:
```suggestion
trunk.  As more transactions are added to the Tangle that reference our transactions – and then more are added that reference those transactions, and so on – this increases the `depth`_ of our transactions.
```

Maybe add a link and/or explanation of depth and how that is used?

docs/transfers.rst
```diff
@@ -0,0 +1,296 @@
...
+:py:meth:`~Iota.send_transfer` will look for ``input addresses`` to fund outgoing
+transactions in the bundle, and auto-generate an unused ``change address`` if
+there is a remainder amount of tokens. It will also take care of finalizing the
+bundle and signing the necessary input transactions.
+
+2. Level Obi-Wan
+~~~~~~~~~~~~~~~~
+Instead of :py:meth:`~Iota.send_transfer`, you can use the combination of
+:py:meth:`~Iota.prepare_transfer` and :py:meth:`~Iota.send_trytes` to achieve
+the same result.
```
todofixthis:
```suggestion
the same result.

.. tip::
  This can be useful if you want to prepare the transactions (including signing inputs) on one device, but you want to then transfer the data to another device for transmission to the Tangle.  For example, you might :py:meth:`~Iota.prepare_transfer` on an air-gapped computer that has your seed stored on it, but then transfer the resulting trytes to a networked computer (that does not have your seed) to :py:meth:`~Iota.send_trytes`.
```

docs/transfers.rst
```diff
@@ -0,0 +1,296 @@
...
+        trytes=prepared_trytes
+    )['trytes']
+
+A difference here is that the end result, ``imaginary_bundle_trytes`` is a list
+of :py:class:`TransactionTrytes`, and not a :py:class:`Bundle` object.
+
+3. Level Yoda
+~~~~~~~~~~~~~
+Being the master Jedi of the PyOTA universe means that you know the most about
+the force of low-level API methods. Use it wisely!
```
todofixthis:
```suggestion
the force of low-level API methods. Use it wisely!

.. tip::
  You generally won't need to split out the process explicitly like this in your application code, but it is useful to understand what :py:meth:`~Iota.send_transfer` does under-the-hood, so that you are better-equipped to troubleshoot any issues that occur.
```

docs/transfers.rst
```diff
@@ -0,0 +1,296 @@
...
+                value=0,
+        )
+    ]
+
+    bundle = ProposedBundle()
+
+    for tx in transactions:
+        bundle.add_transaction(tx)
+
+    # If it was a value transfer, we could
```
todofixthis:
```suggestion
    # If it was a value transfer, we would also need to:
```

docs/transfers.rst
```diff
@@ -0,0 +1,296 @@
...
+    # bundle.add_inputs()
+    # bundle.send_unspent_inputs_to()
+
+    bundle.finalize()
+
+    # Again, for value transfers, we could:
```
todofixthis:
```suggestion
    # Again, for value transfers, we would need to:
```

iota/transaction/base.py
```diff
@@ -570,11 +570,23 @@ def as_tryte_string(self):
             + self.nonce
         )
 
-    def get_signature_validation_trytes(self):
+    def get_bundle_essence_trytes(self):
```
todofixthis: 💭 This is probably low-level enough that it won't break any application logic, but maybe we should make this part of the next minor release (e.g., v2.3.0 => v2.4.0), just to be safe; what do you think?

lzpap: Probably no application uses it, unless it chooses to create the bundle hash itself.
It will be part of the next release, I'm just wondering what is the right approach here. We have in `develop`:
- the async changes, which is pretty big with dropping Python 2  and 3.5 support, and adding the asynchronous api.
- the new networking lib `httpx` with connection pooling is quite faster than `requests`.
- renaming `testnet` to `devnet`, this is a breaking change for apps that relied on it.

Should we include all these in a next minor release 2.4.0, or we say we move to PyOTA 3.0.0 since we dropped py 2 support anyway? :smiley_cat:
Or we wait until summer and reserve PyOTA 3.0.0 for the chrysalis updates?
Probably there will also be some more extended api calls and maybe even account module coming before chrysalis though...

docs/transfers.rst
lzpap: really good point you make here.

docs/transfers.rst
lzpap: It is not, according to https://docs.iota.org/docs/getting-started/0.1/transactions/bundles .

lzpap: Take a look at https://github.com/iotaledger/iri/blob/e1776fbad5d90df86a26402f9025e4b0b2ef7d3e/src/main/java/com/iota/iri/service/API.java#L1256-L1278
The result of this algo is the figure of the bundle structure on the docs site. And btw the documentation on top of the function in IRI is broken I think :smile_cat:

docs/transfers.rst
lzpap: Its branch reference is the other tip, its trunk reference is the same tip as for the other transactions.

docs/transfers.rst
lzpap: thanks for the suggestion! I will come up with something.

iota/transaction/base.py
todofixthis: Ooh, right.  🤔 Dropping Python 2 support is definitely big enough for a major release.  Makes sense to me to bundle it all up in PyOTA 3.

There's no rule that says we have to wait for a long time between major releases; I figure there's no need to make devs wait until Chrysalis, even if that ends up being big enough to justify PyOTA 4 😸

docs/transfers.rst
todofixthis: Oh, that's interesting.  I didn't realise they were flipped 😬

